PR #6, opened 2022-03-14
Optimisations and Fixes

aneojgurhem:


Common/src/gRPC/Services/Submitter.cs
```diff
@@ -276,9 +281,18 @@ await foreach (var taskRequest in taskRequests.WithCancellation(cancellationToke
...
+    activity?.AddTag("sessionId",
+                     session);
+    activity?.AddTag("parentTaskId",
+                     parentTaskId);
+    activity?.AddTag("dispatchId",
+                     dispatchId);
+    activity?.AddTag("taskIds",
+      string.Join(",", requests.Select(request => request.Id)));
+
+    // todo: clean
```
jfonseca-aneo, line 295: Can you add more details to this todo?

Tests/HtcMock/Client/src/GridClient.cs
```diff
@@ -0,0 +1,84 @@
...
+        Id = sessionId,
+      };
+      var session = client_.CreateSession(createSessionRequest);
+      switch (session.ResultCase)
+      {
+        case CreateSessionReply.ResultOneofCase.Error:
+          throw new Exception("Error while creating session : " + session.Error);
+        case CreateSessionReply.ResultOneofCase.None:
+          throw new Exception("Issue with Server !");
+        case CreateSessionReply.ResultOneofCase.Ok:
```
jfonseca-aneo, line 74: a debug log here printing the session id may be useful if in the future a problem within that session arrises.

Common/src/gRPC/Services/Submitter.cs
aneojgurhem: done !

Tests/HtcMock/Client/src/GridClient.cs
aneojgurhem: added

PR #19, opened 2022-03-29
Fix warnings in code

aneojgurhem:


Tests/Stream/Common/TestPayload.cs
```diff
@@ -19,19 +19,19 @@ public enum TaskType
...
+    public string? ResultKey { get; set; }
 
     public byte[] Serialize()
     {
       var jsonString = JsonSerializer.Serialize(this);
       return Encoding.ASCII.GetBytes(StringToBase64(jsonString));
     }
 
-    public static TestPayload? Deserialize(byte[] payload)
+    public static TestPayload? Deserialize(byte[]? payload)
```
jfonseca-aneo, line 34: This nullable type declaration is still relevant after our changes?

aneojgurhem: It should. It does not consider the same kind of payload. This is a object that is serialised and used as payload. It may be interesting to check that this ? is still relevant after our changes.

PR #22, opened 2022-04-03
Fix null payload + add FSM

aneojgurhem:
- Correct [aneoconsulting/ArmoniK#119](https://github.com/aneoconsulting/ArmoniK/issues/119)
- Add Finite State machine to handle the stream more properly

### DIFF
--- a/Adaptors/MongoDB/src/ObjectStorage.cs
+++ b/Adaptors/MongoDB/src/ObjectStorage.cs
@@ -116,7 +116,7 @@ public class ObjectStorage : IObjectStorage
   }
 
   /// <inheritdoc />
-  async IAsyncEnumerable<byte[]> IObjectStorage.TryGetValuesAsync(string key, [EnumeratorCancellation] CancellationToken cancellationToken)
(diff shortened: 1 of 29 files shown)


wkirschenmann (changes requested):
- [x] Rename IObjectStorage.TryGetValuesAsync to IObjectGetValuesAsync end throw KeyNotFoundException instead of yield break.
- [x] Common/src/StateMachines/ComputeRequestStateMachine.cs => remove Register and remove all async
- [x] Common/src/Pollster/ComputeRequestQueue.cs => remove all async
- [ ] Migrate Common/src/Stream/Worker/TaskHandler.cs to the ArmoniK.Core.Common.Tests project
=> https://github.com/aneoconsulting/ArmoniK/issues/123
- [ ] Tests/Stream/Client/StreamWrapperTests.cs => Test EmptyPayload seems very complicated with lots of tests corresponding to known values or expected behavior (ex : switch at line 895). This tests would be simpler if the stream was checked against the submitterService fsm. If this cannot be implemented as part of this PR, create a dedicated issue with corresponding explaination.

wkirschenmann (changes requested):
- [x] Adaptors/MongoDB/src/ObjectStorage.cs : if no chunck is found, throw KeyNotFoundException

wkirschenmann (approved):

PR #43, opened 2022-04-22
Refactor Dispatch

aneojgurhem:
# Refactor Dispatch/RequestProcessor/DataModel/Error Management/Retry

### ResultTable
- Id -> SessionId + Name
- SessionId
- Name
- (Responsible)OwnerTaskId
- Status
    - Created
    - Completed
    - Aborted 
- CreationDate
- Data

Indexes : Id, SessionId, OwnerTaskId, Session + Status

### TaskTable
- SessionId
- ParentTaskIds => List of parents up until first ancestor
- TaskId
- DependencyOutputIds (DataDependencies)
- ExpectedOutputIds
- Status
- StatusMessage
- Options
- CreationDate
- SubmittedDate
- StartDate
- EndDate
- RetryOfIds => List Ids of retried tasks (length is the number of retry), default empty array
- OwnerPodId => stores the Hostname of the pod that processes the task, lock the acquisition of the task by the polling agent
- (PodTTL) => To track pod life, if not updated this means the pod crashed.

### SessionTable
- SessionId
- Options
- Status

## Remove Dispatch
- Dispatch represents an execution of a task and is used to rollback the execution of tasks if one of their parents has an error
- It also does a lock on the execution of a task so that the task is executed by only one worker
- To simplify the rollback of the execution of failed task, the children tasks can be submitted only at the end of the parent task, therefore, there it is easier to cancel the children tasks and rollback the output transfers
- Lock associated to a task will be replaced by a lock on the OwnerPodId in the TaskTable

## Refactor RequestProcessor
- Add FSM for each subpart of the RequestProcessor to process one type of request
- Split the RequestProcessor for each request
- Refactor this part of the code to improve its testablity
```mermaid
graph LR 
    RP["RequestProcessor <br/><br/> data stream (out) <br/> actions stream (in)"] --- WC[WorkerClient]
    WC ---|gRPC| GC[grcpClient]
    GC --- SW[StreamWrapper]
    SW --- APP
    APP["Application <br/><br/> data stream (in) <br/> actions stream (out)"]
```
- Take out the WorkerClient from the RequestProcessor and add a suitable interface for them to interact
- For testing it should be possible to mock the streams directly without worring about gRPC
 
## Pollster
- Use FSM
- PreconditionChecker
- Prefetcher
- RequestProcessor
- DependenciesHandler
- PostProcessor
- DuplicatedMessageHandler

## Lazy Initialization
- List of IInitializable objects
- Initialize the object in the list after dependency injection but before running the application instead of doing an Init in Pollster and Submitter
- Make GrpcServiceSubmitter IInitializable (+everything else we need +Init on everything it needs in its dependency) and use init in Submitter on it when getting it from Dependency Injection

## Refactor ProviderBase
- Hide it better
- Refactor it so that it looks more like a factory
- Make sure that it is easier to test and mock in other classes

## FSM
- Use log scope for each state
- Change scope when there is a transition

## Error / Retry

```mermaid
graph TD
    Task[Task Ends] --> Error{Return Output}
    Error -->|OK| Completed[Task Completed]
    Error -->|KO| Failed
    Error -->|Throw Grpc Error| grpcex{retry < MaxRetries}
    grpcex -->|Yes| SubmitDuplicate[Task Error + Submit task duplicate]
    grpcex -->|No| Failed[Task Failed]
```

- Child tasks submitted at the end of parent execution
- Number of exception in polling agent in health check => if too many, unhealthy
- Task Completed
  - Status Ok
  - Subtask creation OK
  - Outputs OK
  - subtasks are created when received (with creating status i.e; created "on the fly") and submitted at the end of the parent task
- Task Failed
  - Cancellation of outputs and subtasks
  - Status KO and message
  - No retry
  - Managed Error by application (No gRPC Error) but still wants the task to be failed or when the MaxRetries is reached
- Task Error
  - Cancellation of outputs and subtasks
  - Error not managed by application (raise a gRPC Error that will be transfered to the polling agent), annulate output subtasks.
  - create Task Duplicate (with link for monitoring and number of retry) and transfer output creation responsibility

Note : Go closer to gRPC protocol and remove Output. A configuration file should specify the effect of each gRPC return code.

### Task #1
- podA dequeue #1
- podA acquire #1 when OwnerPodId == null
- podA crash
- queue timeout // podA restart
- podB dequeue #1
- podB acquire #1 -> Fail when OwnerPodId != null
- podB check podA for #1 + podA does sanity check on queue
- podB acquire #1 when OwnerPodId == podA and Status != Completed and Status != Completed

## Attention points
- Channel not Disposed in RequestProcessor may cause memory leaks.
- Endpoint on polling agent that returns the current processed task.
- Submit child task that does not have ownership transfer without waiting the end of the parent task.
- Generate taskId by Submitter and return it to the user instead of letting the user generate it.
- Keep in mind the distinction between Result and Output, the former is what the client expects from the task computation, the latter refers to the Status+MessageError.


aneojgurhem:
Next to do :
- refactor request processor
- error management and retries

jfonseca-aneo (approved):

PR #55, opened 2022-05-13
Add several fully integrated tests and ignore full test for retry due…

aneojgurhem:
… to poor error management in pollster

### DIFF
--- a/Common/src/Pollster/RequestProcessor.cs
+++ b/Common/src/Pollster/RequestProcessor.cs
@@ -238,6 +238,10 @@ public class RequestProcessor : IDisposable
 
     workerStreamHandler_.StartTaskProcessing(taskData,
                                              cancellationToken);
+    if (workerStreamHandler_.Pipe is null)
+    {
+      throw new ArmoniKException($"{nameof(IWorkerStreamHandler.Pipe)} should not be null");
+    }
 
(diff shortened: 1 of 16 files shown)


Common/src/Pollster/RequestProcessor.cs
```diff
@@ -333,30 +334,29 @@ await output.WhenAll()
...
                                                                     {
-                                                                      CreateTask = new ProcessRequest.Types.CreateTask
-                                                                                   {
-                                                                                     Reply   = replyLargeTasksAsync,
-                                                                                     ReplyId = first.RequestId,
-                                                                                   },
+                                                                      Reply   = replyLargeTasksAsync,
+                                                                      ReplyId = first.RequestId,
                                                                     },
-                                                                    CancellationToken.None)
```
ddubuc, line 344: No more CancellationToken for this switch case ?

ddubuc (commented):
Is there any issue with no cancellationToken in CreateTask ?

Common/src/Pollster/RequestProcessor.cs
aneojgurhem: We made an interface above this that wrap up a grpc type that has this cancellation token in its implementation (grpcAsyncPipe). It has a None cancellation token because its implementation does not support a cancellation token.

ddubuc (approved):

PR #59, opened 2022-05-17
Fix cancel session

aneojgurhem:


Adaptors/MongoDB/src/TaskTable.cs
```diff
@@ -241,7 +241,8 @@ public class TaskTable : ITaskTable
     var taskCollection = taskCollectionProvider_.Get();
 
     // method is final for taskStatus
-    var result = await taskCollection.UpdateManyAsync(model => model.SessionId == sessionId,
+    var result = await taskCollection.UpdateManyAsync(model => model.SessionId == sessionId         && model.Status != TaskStatus.Completed &&
+                                                               model.Status    != TaskStatus.Failed && model.Status != TaskStatus.Canceled,
```
ngruelaneo, line 245: Why are you checking that the status is at Canceled? Is it possible to call this function multiple time to apply this status?

aneojgurhem: These are the 3 final status. Checking them like that allows to do it in only one request

ngruelaneo (approved):

PR #60, opened 2022-05-18
Make sure that there is a connection to the worker before acquiring t…

aneojgurhem:
…asks in polling agent


Common/src/Stream/Worker/WorkerStreamHandler.cs
```diff
@@ -82,14 +85,44 @@ public void StartTaskProcessing(TaskData taskData, CancellationToken cancellatio
...
+        isInitialized_ = true;
+        logger_.LogInformation("Channel was initialized");
+        return;
+      }
+      catch (Exception ex)
+      {
+        logger_.LogDebug(ex,
+                         "Channel was not created, retry in {seconds}s",
+                         retry * retry);
+        await Task.Delay(1000 * retry * retry,
```
ngruelaneo, line 116: Parametrize the delay

wkirschenmann (approved):

PR #63, opened 2022-05-30
Consider not implemented response for worker health check as OK in po…

aneojgurhem:
…lling agent


Common/src/Stream/Worker/WorkerStreamHandler.cs
```diff
@@ -113,6 +113,12 @@ public async Task Init(CancellationToken cancellationToken)
         logger_.LogInformation("Channel was initialized");
         return;
       }
+      catch (RpcException ex) when (ex.StatusCode == StatusCode.Unimplemented)
+      {
+        isInitialized_ = true;
+        logger_.LogInformation("Channel was initialized");
```
ngruelaneo, line 119: add in the logger that you went trough an "unimplemented" exception.

ngruelaneo (approved):

PR #69, opened 2022-06-07
Simplify and factorize code of stream test

aneojgurhem:


Tests/Stream/Client/SubmitterExt.cs
```diff
@@ -0,0 +1,65 @@
...
+using ArmoniK.Api.gRPC.V1;
+using ArmoniK.Core.Common.Stream.Client;
+
+using Google.Protobuf.WellKnownTypes;
+
+namespace ArmoniK.Extensions.Common.StreamWrapper.Tests.Client;
+
+public static class SubmitterExt
+{
+  public static void CreateSessionExt(this Submitter.SubmitterClient client,
```
wkirschenmann, line 14: The class name already means that this is an extension method. This one should be named `CreateSession`. It will appear as an overload of the existing `CreateSession` methods

wkirschenmann (changes requested):

wkirschenmann (approved):

PR #73, opened 2022-06-09
Fix for Amqp idle time out 

ddubuc:


Adaptors/Amqp/src/SessionAmqp.cs
```diff
@@ -0,0 +1,133 @@
...
+                                                                        default:
+                                                                          Logger.LogError("SSL error : {error}",
+                                                                                          errors);
+                                                                          return false;
+                                                                      }
+                                                                    };
+      }
+
+      var retries = 1;
+      while (true)
```
wkirschenmann, line 89: Avoid 'while(true)' 
Use 'for(Var retry=0; retry < maxRetry ; ++retry)' instead

Adaptors/Amqp/src/SessionAmqp.cs
```diff
@@ -0,0 +1,133 @@
...
+      else
+      {
+        logger.LogWarning("AMQP Connection closed with error: {0}",
+                        error.ToString());
+      }
+    }
+
+    public async Task<SessionAmqp> OpenConnection()
+    {
+      if (Session != null)
```
wkirschenmann, line 50: Shouldn't we check if the session is not already open before closing it?

Adaptors/Amqp/src/SessionAmqp.cs
```diff
@@ -0,0 +1,133 @@
...
+
+    public SessionAmqp(Options.Amqp options,
+                       ILogger      logger)
+    {
+      Options = options;
+      Logger  = logger;
+
+    }
+
+    private int retriesReconnect_ = 5;
```
wkirschenmann, line 33: Max value should be defined in options

wkirschenmann (changes requested):

Adaptors/Amqp/src/SessionAmqp.cs
```diff
@@ -0,0 +1,161 @@
...
+    retriesReconnect_ = options.MaxRetry;
+  }
+
+  public Session      Session { get; set; }
+  public Options.Amqp Options { get; set; }
+
+  public ILogger Logger { get; set; }
+
+  private static void OnCloseConnection(IAmqpObject sender,
+                                        Error       error,
```
lemaitre-aneo, line 58: Would it be better to use `error?`

Adaptors/Amqp/src/SessionAmqp.cs
```diff
@@ -0,0 +1,161 @@
...
+      }
+      catch
+      {
+        Thread.Sleep(1000 * retry);
+      }
+    }
+
+    if (retry == Options.MaxRetry)
+    {
+      throw new ArgumentOutOfRangeException(nameof(Options.MaxRetry),
```
lemaitre-aneo, line 135: I think `Timeout` is a better name. I don't know if there already is a dotnet exception like that, but I guess so.

lemaitre-aneo (dismissed):

Adaptors/Amqp/src/SessionAmqp.cs
```diff
@@ -55,7 +57,7 @@ public class SessionAmqp
   public ILogger Logger { get; set; }
 
   private static void OnCloseConnection(IAmqpObject sender,
-                                        Error       error,
+                                        [CanBeNull] Error      error,
```
lemaitre-aneo, line 60: Use `?` instead. You might need to change the project settings for that.

docker-compose/docker-compose-htcmock/docker-compose.yml
```diff
@@ -24,6 +24,7 @@ services:
       - Amqp__MaxPriority=10
       - Amqp__User=admin
       - Amqp__Password=admin
+      - Amqp__MaxRetry=5
```
lemaitre-aneo, line 27: Couldn't you add this configuration in the appsettings? this would allow to have a default value and make the current infrastructure compatible with this change.

jfonseca-aneo: Yes, I plan to do so

lemaitre-aneo (dismissed):

lemaitre-aneo (approved):

wkirschenmann (approved):

PR #79, opened 2022-06-16
Add TaskHandler to manage task lifecycle

aneojgurhem:
Preparations for reworking task acquisition and improve management of errors in agent


Common/src/Pollster/Pollster.cs
```diff
@@ -120,8 +124,6 @@ await Init(cancellationToken)
         await foreach (var message in messages.WithCancellation(cancellationToken)
```
lemaitre-aneo, line 124: Why `message` is not "used" anymore (ie: `using`). When is it disposed?

jfonseca-aneo (approved):

Common/src/Pollster/Pollster.cs
aneojgurhem: It is explicitly disposed in the TaskHandler Dispose function

PR #98, opened 2022-06-28
Set task to error when there is an issue during task resubmission

aneojgurhem:
fix https://github.com/aneoconsulting/ArmoniK/issues/304

### DIFF
--- a/Tests/HtcMock/Client/src/SessionClient.cs
+++ b/Tests/HtcMock/Client/src/SessionClient.cs
@@ -136,8 +136,7 @@ public class SessionClient : ISessionClient
 
     foreach (var (payload, dependencies) in payloadsWithDependencies)
     {
-      var taskId = Guid.NewGuid()
-                       .ToString();
+      var taskId = "root";
       logger_.LogDebug("Create task {taskId}",
                        taskId);
       var taskRequest = new TaskRequest
--- a/Tests/HtcMock/Server/src/SampleComputerService.cs
+++ b/Tests/HtcMock/Server/src/SampleComputerService.cs
@@ -26,6 +26,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Text;
+using System.Threading;
 using System.Threading.Tasks;
 
 using ArmoniK.Api.gRPC.V1;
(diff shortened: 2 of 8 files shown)


Tests/HtcMock/Client/src/SessionClient.cs
```diff
@@ -136,8 +136,7 @@ public IEnumerable<string> SubmitTasksWithDependencies(IEnumerable<Tuple<byte[],
 
     foreach (var (payload, dependencies) in payloadsWithDependencies)
     {
-      var taskId = Guid.NewGuid()
-                       .ToString();
+      var taskId = "root";
```
jfonseca-aneo, line 139: Won't this change imply mutiple subtasks with identical ids in one session?

lemaitre-aneo: Yes, that is intended to produce more reliably the error we try to recover from.

Tests/HtcMock/Server/src/SampleComputerService.cs
```diff
@@ -189,25 +163,69 @@ public override async Task<Output> Process(ITaskHandler taskHandler)
...
+             };
     }
-    catch (Exception ex)
+    catch (RpcException ex)
     {
+      var taskRpcException = taskHandler.TaskOptions.GetValueOrDefault("TaskRpcException",
+                                                                       string.Empty);
+      if (taskRpcException != string.Empty && taskHandler.TaskId.EndsWith(taskRpcException))
+      {
+        Thread.Sleep(TimeSpan.FromSeconds(15));
```
lemaitre-aneo, line 201: What is the purpose of this timeout?

Tests/HtcMock/Client/src/SessionClient.cs
aneojgurhem: In the future, we want that the task id generated by the submitter so it will fix this issue

Tests/HtcMock/Server/src/SampleComputerService.cs
aneojgurhem: It was a test. It is not necessary. I will remove it.

Tests/HtcMock/Server/src/SampleComputerService.cs
```diff
@@ -26,6 +26,7 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Text;
+using System.Threading;
```
lemaitre-aneo, line 29: Not needed anymore?

lemaitre-aneo (approved):

lemaitre-aneo (approved):

PR #104, opened 2022-07-05
Implement split of agent/worker bidirectionnal stream

aneojgurhem:


Common/src/Pollster/AgentHandler.cs
```diff
@@ -0,0 +1,108 @@
...
+
+    var builder = WebApplication.CreateBuilder();
+
+    builder.Services.AddSingleton(agent_)
+           .AddLogging(loggerInit_.Configure)
+           .AddGrpc();
+
+    builder.WebHost.ConfigureKestrel(options =>
+                                     {
+                                       if (File.Exists(socketPath))
```
lemaitre-aneo, line 65: This is useless and will hide issues.

aneojgurhem: the deletion of the file ?

lemaitre-aneo: Yes. On Linux, you can delete a file in use by another process, and the other process will keep the file opened nonetheless, but the file will no longer be accessible by its path.
Therefore, when a new socket will be created, a new file with a new inode will be created, and you will be able to bind to it even in the case where another process was still listening on the previous socket.
Without deleting the socket file, I expect the server to crash when it tries to listen to it while another process is still listening on it.

Common/src/Pollster/AgentHandler.cs
```diff
@@ -0,0 +1,142 @@
...
+
+      var builder = WebApplication.CreateBuilder();
+
+      builder.Host.UseSerilog(loggerInit.GetSerilogConf());
+
+      builder.Services.AddLogging(loggerInit.Configure)
+             .AddSingleton<GrpcAgentService>()
+             .AddGrpc();
+
+      builder.WebHost.ConfigureKestrel(options => options.ListenUnixSocket(computePlanOptions.AgentChannel.Address!,
```
ngruelaneo, line 79: Stupid question: what is a Kestrel ?

Common/src/Stream/Worker/TaskRequestExtensions.cs
```diff
@@ -40,27 +41,27 @@ namespace ArmoniK.Core.Common.Stream.Worker;
...
     var fsm = new ProcessReplyCreateLargeTaskStateMachine(NullLogger.Instance);
     fsm.InitRequest();
     if (taskOptions is not null)
     {
-      yield return new ProcessReply.Types.CreateLargeTaskRequest
+      yield return new CreateTaskRequest
                    {
-                     InitRequest = new ProcessReply.Types.CreateLargeTaskRequest.Types.InitRequest
+                     InitRequest = new CreateTaskRequest.Types.InitRequest
                                    {
```
ngruelaneo, line 55: Stupid question: {}and () are equivalent in c#?

Common/src/Stream/Worker/TaskRequestExtensions.cs
```diff
@@ -40,27 +41,27 @@ namespace ArmoniK.Core.Common.Stream.Worker;
...
                                    },
                    };
     }
     else
     {
-      yield return new ProcessReply.Types.CreateLargeTaskRequest
+      yield return new CreateTaskRequest
                    {
-                     InitRequest = new ProcessReply.Types.CreateLargeTaskRequest.Types.InitRequest(),
+                     InitRequest        = new CreateTaskRequest.Types.InitRequest(),
```
ngruelaneo, line 64: why?  These spaces/tabulations are not event coherent with the line 54.

Common/src/Stream/Worker/WorkerStreamHandler.cs
```diff
@@ -73,20 +67,20 @@ public Queue<ComputeRequest> WorkerReturn()
       throw new ArmoniKException("Worker client should be initialized");
     }
 
-    stream_ = workerClient_.Process(deadline: DateTime.UtcNow + taskData.Options.MaxDuration,
+
+    // todo : using var ? stream is IDisposable
```
ngruelaneo, line 71: comments with a todo.

Common/src/Stream/Worker/WorkerStreamHandler.cs
```diff
@@ -73,20 +67,20 @@ public Queue<ComputeRequest> WorkerReturn()
...
+    var stream = workerClient_.Process(deadline: DateTime.UtcNow + taskData.Options.MaxDuration,
                                     cancellationToken: cancellationToken);
 
-    if (stream_ is null)
+    if (stream is null)
```
ngruelaneo, line 75: why are you using the operator `is` here and at other place `=`?

Common/src/Stream/Worker/WorkerStreamWrapper.cs
```diff
@@ -47,38 +47,31 @@ public WorkerStreamWrapper(ILoggerFactory loggerFactory)
...
-                        .ConfigureAwait(false);
-    if (await requestStream.MoveNext(context.CancellationToken)
-                           .ConfigureAwait(false))
+    Output output;
     {
-      throw new InvalidOperationException("The request stream is expected to be finished.");
+      await using var taskHandler = await TaskHandler.Create(requestStream,
+                                                             new Configuration
+                                                             {
+                                                               DataChunkMaxSize = 50 * 1024,
```
ngruelaneo, line 58: will this MaxSize will never change? Why not use a parameter?

Common/src/gRPC/Services/GrpcAgentService.cs
```diff
@@ -0,0 +1,130 @@
...
+    if (agent_ != null)
+    {
+      return await agent_.CreateTask(requestStream,
+                                     context.CancellationToken)
+                         .ConfigureAwait(false);
+    }
+
+    return new CreateTaskReply
+           {
+             Error = "No task is accepting request",
```
ngruelaneo, line 67: I suppose that the Error variable is there so you can obtain a specific message if there are an error?

Common/src/Pollster/AgentHandler.cs
wkirschenmann: > Kestrel is a cross-platform web server for ASP.NET Core. Kestrel is the web server that's included and enabled by default in ASP.NET Core project templates.
>
> Kestrel supports the following scenarios:
> 
> - HTTPS 
> - [HTTP/2](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/servers/kestrel/http2?view=aspnetcore-6.0) (except on macOS†)
> - Opaque upgrade used to enable [WebSockets](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/websockets?view=aspnetcore-6.0)
> - Unix sockets for high performance behind Nginx
> 
> †HTTP/2 will be supported on macOS in a future release.
>
>Kestrel is supported on all platforms and versions that .NET Core supports.

https://docs.microsoft.com/en-us/aspnet/core/fundamentals/servers/kestrel?view=aspnetcore-6.0

Tests/HtcMock/Server/src/SampleComputerService.cs
```diff
@@ -122,7 +122,7 @@ public override async Task<Output> Process(ITaskHandler taskHandler)
         await taskHandler.CreateTasksAsync(readyRequests.Select(r =>
                                                                 {
                                                                   var taskId = taskHandler.SessionId + "%" + r.Id;
-                                                                  logger_.LogDebug("Create task {taskId}",
+                                                                  logger_.LogDebug("Create task {task}",
```
ngruelaneo, line 125: Reading on github is not easy sometimes... Where is the variable `task`?

Common/src/Stream/Worker/WorkerStreamHandler.cs
wkirschenmann: Most of the time, they are equivalent. However, in some corner cases, differences may appear. For instance if one class overloads the equality operator. 
Also, when testing types, `is` will check if the object either inherits from the given types or if it implements the given interface.

https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/is

Common/src/Stream/Worker/WorkerStreamWrapper.cs
wkirschenmann: The value can be hardcoded but it needs to be set up in only one place commonly accessible from everywhere. One good approach is to have it in a `Constants` public static class containing public const values.

An even better approach would be for the worker to obtain this value from the polling agent.

Common/src/Pollster/DataPrefetcher.cs
```diff
@@ -84,13 +85,15 @@ await objectStorageFactory_.Init(cancellationToken)
...
   /// <param name="taskData">Task metadata</param>
+  /// <param name="socketPath">Path to the socket used for receiving the requests from the worker</param>
   /// <param name="cancellationToken">Token used to cancel the execution of the method</param>
   /// <returns>
   /// Queue containing the request containing the data for the task which can be sent to the worker
   /// </returns>
   /// <exception cref="ObjectDataNotFoundException">input data are not found</exception>
   /// <exception cref="InvalidOperationException">invalid transition between states</exception>
   public async Task<Queue<ProcessRequest.Types.ComputeRequest>> PrefetchDataAsync(TaskData          taskData,
+                                                                                  string            socketPath,
```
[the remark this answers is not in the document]
aneojgurhem: According to defined formatting rules, code is appropriately formatted.

Common/src/Pollster/AgentHandler.cs
```diff
@@ -0,0 +1,142 @@
...
+    }
+    catch (Exception e)
+    {
+      logger.LogError(e,
+                      "Error while initializing agent server");
+      throw;
+    }
+  }
+
+  public async Task Start(IAgent            agent,
```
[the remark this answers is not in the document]
aneojgurhem: You're rigth, the following lines are not formatted correctly

```diff
-                          string token,
-                          ILogger logger,
+                          string            token,
+                          ILogger           logger,
```

Common/src/Stream/Worker/TaskRequestExtensions.cs
aneojgurhem: It depends on what you want to do.
Usually, {} is a kind of rapid initialization of the internal data whereas () is the constructor.

Common/src/Stream/Worker/TaskRequestExtensions.cs
aneojgurhem: Formatting issue^^ A way to check if the code is properly formatted has been merged on master. We will be enforcing it soon.

Common/src/Stream/Worker/WorkerStreamWrapper.cs
aneojgurhem: There is a class that holds this value. I think I will send it to the worker via the compute request. It has always been there.

Common/src/gRPC/Services/GrpcAgentService.cs
aneojgurhem: Yes

Common/src/Stream/Worker/TaskRequestExtensions.cs
wkirschenmann: As a first approximation, 
```
var t = new T
{
  Toto = XXX;
};
```
is the same as:
```
var t = new T();
t.Toto = XXX;
```

For more details: https://docs.microsoft.com/en-us/cpp/cpp/initializing-classes-and-structs-without-constructors-cpp?view=msvc-170

Tests/HtcMock/Server/src/SampleComputerService.cs
aneojgurhem: It is not a variable, it is a way to name a variable in logs so that it is searchable. The actual value is passed as another argument.

Common/src/Stream/Worker/WorkerStreamHandler.cs
aneojgurhem: done

Common/src/Pollster/IAgentHandler.cs
```diff
@@ -0,0 +1,57 @@
...
+public interface IAgentHandler
+{
+  /// <summary>
+  /// Stops the handler
+  /// </summary>
+  /// <param name="cancellationToken">Token used to cancel the execution of the method</param>
+  /// <returns>
+  /// Task representing the asynchronous execution of the method
+  /// </returns>
+  Task Stop(CancellationToken cancellationToken);
```
lemaitre-aneo, line 41: I would have thought something like this:
```c#
public async Task ReleaseServer(WebApplication, ...);
public async Task<WebApplication> GetServer(...);
```

Common/src/Pollster/TaskHandler.cs
```diff
@@ -84,7 +91,9 @@ internal class TaskHandler : IAsyncDisposable
     workerStreamHandler_   = workerStreamHandler;
     objectStorageFactory_  = objectStorageFactory;
     activitySource_        = activitySource;
+    agentHandler_          = agentHandler;
     logger_                = logger;
+    socketPath_       = socketPath;
```
lemaitre-aneo, line 96: Formatting typo

Common/src/Pollster/TaskHandler.cs
```diff
@@ -382,18 +393,40 @@ public async Task ExecuteTask(CancellationToken cancellationToken)
...
-                                           .ConfigureAwait(false);
+    logger_.LogDebug("Create agent server to receive requests from worker");
+
+    agent_ = new Agent(submitter_,
+                           objectStorageFactory_,
+                           taskData_.SessionId,
+                           taskData_.TaskId,
+                           logger_);
+
+    await agentHandler_.Start(agent_,
```
lemaitre-aneo, line 404: You can start the agent before, and only do the await here.

Common/src/Pollster/TaskHandler.cs
```diff
@@ -408,21 +441,98 @@ public async Task PostProcessing(CancellationToken cancellationToken)
...
   {
     using var _ = logger_.BeginNamedScope("DisposeAsync",
                                           ("taskId", messageHandler_.TaskId));
+
+    logger_.LogDebug("MessageHandler status is {status}",
+                     messageHandler_.Status);
     await messageHandler_.DisposeAsync()
                          .ConfigureAwait(false);
+
+    agent_?.Dispose();
```
lemaitre-aneo, line 536: `agentHandler_.Stop(agent_)`?

Common/src/Stream/Worker/TaskHandler.cs
```diff
@@ -104,54 +143,21 @@ public IList<string> ExpectedResults
...
-      {
-        throw new AggregateException(reply.NonSuccessfullIds.Ids.Select(s => new InvalidOperationException($"Could not create task it id={s}")));
-      }
+      await stream.RequestStream.WriteAsync(createLargeTaskRequest,
+                                            CancellationToken.None)
+                  .ConfigureAwait(false);
     }
-    finally
+
+    var reply = await stream.ResponseAsync.ConfigureAwait(false);
```
lemaitre-aneo, line 157: `.configureAwait(false)` is not on its own line.
/s

Common/src/Stream/Worker/TaskRequestExtensions.cs
```diff
@@ -40,27 +41,27 @@ namespace ArmoniK.Core.Common.Stream.Worker;
 [PublicAPI]
 public static class TaskRequestExtensions
 {
-  public static IEnumerable<ProcessReply.Types.CreateLargeTaskRequest> ToRequestStream(this IEnumerable<TaskRequest> taskRequests,
-                                                                                       TaskOptions?                  taskOptions,
-                                                                                       int                           chunkMaxSize)
+  public static IEnumerable<CreateTaskRequest> ToRequestStream(this IEnumerable<TaskRequest> taskRequests,
```
lemaitre-aneo, line 44: `IAsyncEnumerable<...>`?

Common/src/Stream/Worker/TaskRequestExtensions.cs
```diff
@@ -99,13 +100,13 @@ public static class TaskRequestExtensions
     }
   }
 
-  public static IEnumerable<ProcessReply.Types.CreateLargeTaskRequest> ToRequestStream(this TaskRequest                        taskRequest,
-                                                                                       bool                                    isLast,
-                                                                                       int                                     chunkMaxSize,
-                                                                                       ProcessReplyCreateLargeTaskStateMachine processReplyCreateLargeTaskStateMachine)
+  public static IEnumerable<CreateTaskRequest> ToRequestStream(this TaskRequest                        taskRequest,
```
lemaitre-aneo, line 103: `IAsyncEnumerable<...>`?

Common/src/gRPC/GrpcChannelProvider.cs
```diff
@@ -40,19 +40,22 @@
 namespace ArmoniK.Core.Common.gRPC;
 
 [UsedImplicitly]
-public class GrpcChannelProvider
+public class GrpcChannelProvider : IDisposable
```
lemaitre-aneo, line 43: `IAsyncDisposable`?

Common/src/gRPC/Services/Agent.cs
```diff
@@ -0,0 +1,416 @@
...
+
+using Result = ArmoniK.Api.gRPC.V1.Agent.Result;
+
+namespace ArmoniK.Core.Common.gRPC.Services;
+
+
+/// <summary>
+/// Represents the internal processing requests received by the agent. Provides methods to process those requests
+/// </summary>
+public class Agent : IAgent, IDisposable
```
lemaitre-aneo, line 47: `IAsyncDisposable`?

Common/tests/FullIntegration/WorkerStreamHandlerErrorRetryTest.cs
```diff
@@ -54,8 +55,8 @@ public WorkerStreamHandlerErrorRetryTest(Exception exception)
 
     var task = new Task(async () =>
                         {
-                          var request = await ChannelAsyncPipe.Reverse.Reader.FirstAsync(cancellationToken: cancellationToken)
-                                                   .ConfigureAwait(false);
+                          var request = await ChannelAsyncPipe.Reverse.Read(cancellationToken)
```
lemaitre-aneo, line 58: is `Read` implicitely async ?

Common/src/Stream/Worker/WorkerStreamWrapper.cs
aneojgurhem: The configuration was already sent but not used. Now it is

Common/src/Stream/Worker/TaskHandler.cs
aneojgurhem: Appropriatly formatted according to the given rules

Common/tests/Pollster/DataPrefetcherTest.cs
```diff
@@ -393,6 +396,7 @@ public async Task EmptyPayloadAndOneDependencyWithDataStateMachine()
                                                                   DateTime.Now,
                                                                   new Output(true,
                                                                              "")),
+                                                     "/socket",
```
jfonseca-aneo, line 399: inconsistent formating

Common/src/Pollster/IAgentHandler.cs
```diff
@@ -0,0 +1,57 @@
+﻿// This file is part of the ArmoniK project
+//
+// Copyright (C) ANEO, 2021-$CURRENT_YEAR$. All rights reserved.
```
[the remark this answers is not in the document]
aneojgurhem: In another PR, it seems that there is an issue with the template

Common/tests/Pollster/DataPrefetcherTest.cs
aneojgurhem: Actually, it is the right one !

jfonseca-aneo: Ok; the ones above are not formated like this

aneojgurhem: My bad, there was a mistake there. It is fixed now

Common/src/Pollster/TaskHandler.cs
```diff
@@ -408,21 +446,98 @@ public async Task PostProcessing(CancellationToken cancellationToken)
   {
     using var _ = logger_.BeginNamedScope("PostProcessing",
                                           ("taskId", messageHandler_.TaskId));
-    if (processResult_ == null)
+
+    if (workerStreamHandler_.Pipe == null || taskData_ == null || agent_ == null)
     {
-      throw new NullReferenceException();
+      throw new ArgumentNullException();
```
jfonseca-aneo, line 452: update description in the doc to new returned exception

Common/src/Pollster/AgentHandler.cs
```diff
@@ -0,0 +1,145 @@
...
+      builder.WebHost.ConfigureKestrel(options => options.ListenUnixSocket(computePlanOptions.AgentChannel.Address!,
+                                                                           listenOptions => listenOptions.Protocols = HttpProtocols.Http2));
+
+      app_ = builder.Build();
+
+      app_.UseRouting();
+      app_.MapGrpcService<GrpcAgentService>();
+
+      service_ = app_.Services.GetRequiredService<GrpcAgentService>();
+      app_.Start();
```
lemaitre-aneo, line 89: Would it be possible to move the constructor code into a factory function in order to have it asynchronous?

lemaitre-aneo (approved):

Common/src/Pollster/AgentHandler.cs
aneojgurhem: We would like to implement a better way to distinguish between Dependency Injection and actual asynchronous initialization of objects.- This could be done at this point.

PR #105, opened 2022-07-07
Check format

lemaitre-aneo:
Add a step to the github action to check the format of the code.
For now, the check can fail. As soon as the repo pass this test, we can enable failures.


.editorconfig
```diff
@@ -0,0 +1,148 @@
+root = true
+
+[*.cs]
+indent_style=space
+indent_size=2
+tab_width=2
+end_of_line=lf
+charset=utf-8
+trim_trailing_whitespace=true
```
aneojgurhem, line 9: This one can be put everywhere as well as :
```
end_of_line=lf
charset=utf-8
```

lemaitre-aneo: done

aneojgurhem (approved):

PR #111, opened 2022-07-21
Migrate Stream Wrapper to ArmoniK.Api

aneojgurhem:


ngruelaneo (approved):

lemaitre-aneo (commented):
Would have been cool to not have formatting changes in this PR... Apart from that, all is good.

aneojgurhem:
The formatting changes come from the fact that those files were modified during this pr

.gitmodules
```diff
@@ -1,3 +0,0 @@
-[submodule "Api"]
-	path = Api
-	url = https://github.com/aneoconsulting/ArmoniK.Api.git
```
jfonseca-aneo, line 3: the submodule is only required for the StreamWrapper?

aneojgurhem: No, but I made packages in ArmoniK.Api. Therefore, I used them instead of the submodule

jfonseca-aneo: So if we remove it as submodule in Core we need to document the requirement (in case someone clones the repo)

aneojgurhem: The package is included in the .csproj so this makes it a requirement already.

jfonseca-aneo: Ah ok!

Tests/HtcMock/Client/src/ArmoniK.Samples.HtcMock.Client.csproj
```diff
@@ -24,16 +24,18 @@
 	</PropertyGroup>
 
 	<ItemGroup>
+	  <PackageReference Include="ArmoniK.Api.Client" Version="3.0.0-jgmigrate.81.e1930f1" />
```
jfonseca-aneo, line 27: we need a proper tag here

jfonseca-aneo (approved):

PR #114, opened 2022-07-25
Partition in Sessions and tasks

lemaitre-aneo:


Common/src/gRPC/Services/Submitter.cs
```diff
@@ -133,27 +133,50 @@ public class Submitter : ISubmitter
...
-    options ??= await sessionTable_.GetDefaultTaskOptionAsync(sessionId,
-                                                              cancellationToken)
-                                   .ConfigureAwait(false);
+    options = options != null ? ArmoniK.Core.Common.Storage.TaskOptions.Merge(options, sessionData.Options) : sessionData.Options;
+
+    var partitionId = options.PartitionId;
+    var availablePartitionIds = sessionData.PartitionIds ?? Array.Empty<string>();
+    if (!availablePartitionIds.Contains(partitionId))
+    {
+      throw new InvalidOperationException($"The session {sessionData.SessionId} is assigned to the partitions " +
```
jfonseca-aneo, line 177: add a test in the SubmitterTests.cs that runs into this exception

Adaptors/Memory/src/SessionTable.cs
```diff
@@ -59,17 +59,31 @@ public Task Init(CancellationToken cancellationToken)
...
+  /// <inheritdoc />
+  public Task<SessionData> GetSession(string            sessionId,
+                                      CancellationToken cancellationToken = default)
+  {
+    if (!storage_.ContainsKey(sessionId))
+    {
+      throw new SessionNotFoundException($"Key '{sessionId}' not found");
+    }
+
+    return Task.FromResult(storage_[sessionId]);
```
ngruelaneo, line 84: What is happenning if sessionId is not found?

Common/src/Storage/SessionData.cs
```diff
@@ -32,15 +33,18 @@ public record SessionData(string        SessionId,
                           SessionStatus Status,
                           DateTime      CreationDate,
                           DateTime?     CancellationDate,
+                          IList<string> PartitionIds,
```
ngruelaneo, line 36: Why using a Ilist here and at some other place IEnumerable?

Common/src/gRPC/Services/Submitter.cs
```diff
@@ -133,27 +133,50 @@ public class Submitter : ISubmitter
...
+  public async Task<(IEnumerable<Storage.TaskRequest> requests, int priority)> CreateTasks(SessionData                   sessionData,
+                                                                                           string                        parentTaskId,
+                                                                                           TaskOptions                   options,
+                                                                                           IAsyncEnumerable<TaskRequest> taskRequests,
+                                                                                           CancellationToken             cancellationToken)
   {
     using var logFunction = logger_.LogFunction(parentTaskId);
     using var activity    = activitySource_.StartActivity($"{nameof(CreateTasks)}");
-    using var sessionScope = logger_.BeginPropertyScope(("Session", sessionId),
+    using var sessionScope = logger_.BeginPropertyScope(("Session", sessionData.SessionId),
```
ngruelaneo, line 163: The partition can be deduce from the Task?

Adaptors/Memory/src/SessionTable.cs
lemaitre-aneo: An exception is throwed a few lines earlier

Common/src/Storage/SessionData.cs
lemaitre-aneo: I think it should even be `string[]`
@wkirschenmann opinion?

lemaitre-aneo: kept IList

Common/src/gRPC/Services/Submitter.cs
lemaitre-aneo: added PartitionId to scope

Common/tests/Submitter/SubmitterTests.cs
```diff
@@ -310,6 +312,32 @@ public async Task CreateSessionShouldSucceed()
...
+    var defaultTaskOptions = new TaskOptions
+                             {
+                               MaxDuration = Duration.FromTimeSpan(TimeSpan.FromSeconds(2)),
+                               MaxRetries  = 2,
+                               Priority    = 1,
+                               PartitionId = "invalid",
+                             };
+
+    /*
+    // TODO: Once partitions are fully integrated, creating a session without partition should fail
```
jfonseca-aneo, line 327: One can avoid the commented code by adding the attribute [Ignore("reason to ignore the test")]  to the test

Adaptors/Memory/src/SessionTable.cs
```diff
@@ -59,28 +60,38 @@ public Task Init(CancellationToken cancellationToken)
...
+  }
+
+  /// <inheritdoc />
+  public Task<bool> IsSessionCancelledAsync(string            sessionId,
+                                            CancellationToken cancellationToken = default)
+  {
+    return Task.FromResult(this.GetSessionAsync(sessionId,
+                                                cancellationToken)
+                               .Result.Status == SessionStatus.Canceled);
```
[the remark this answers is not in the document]
lemaitre-aneo: `.Result` does not wait. The only reason it does work here is because Task is created already completed. This is a safe assumption because the call is to another method of the same class (same file), and even just a few lines above.

Common/src/Utils/EnumerableExt.cs
```diff
@@ -1,33 +1,37 @@
-﻿// This file is part of the ArmoniK project
-// 
-// Copyright (C) ANEO, 2021-2022. All rights reserved.
+// This file is part of the ArmoniK project
+//
+// Copyright (C) ANEO, 2021-$CURRENT_YEAR$. All rights reserved.
```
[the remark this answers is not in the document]
lemaitre-aneo: Apart from doing this manually, do you know how to solve this?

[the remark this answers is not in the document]
jfonseca-aneo: Unfortunately I don't

jfonseca-aneo (approved):

PR #117, opened 2022-08-03
Reformat everything

lemaitre-aneo:
Also makes the pipeline fail if the formatting is not correct.


jfonseca-aneo (approved):

jfonseca-aneo (approved):

Adaptors/Memory/src/ArmoniK.Core.Adapters.Memory.csproj
```diff
@@ -9,6 +9,7 @@
     <PackageRequireLicenseAcceptance>True</PackageRequireLicenseAcceptance>
     <GeneratePackageOnBuild>True</GeneratePackageOnBuild>
     <IsPackable>true</IsPackable>
+    <Nullable>enable</Nullable>
```
aneojgurhem, line 12: Why is this here ? It's reformatting pr !

lemaitre-aneo: Because it forces resharper to use `?` instead of `[CanBeNull]`.

PR #125, opened 2022-08-11
Fix metrics exporter

aneojgurhem:
fix https://github.com/aneoconsulting/ArmoniK/issues/428

### DIFF
--- a/Control/Metrics/src/ArmoniKMeter.cs
+++ b/Control/Metrics/src/ArmoniKMeter.cs
@@ -25,6 +25,7 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics.Metrics;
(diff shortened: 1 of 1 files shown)


jfonseca-aneo (dismissed):

Control/Metrics/src/ArmoniKMeter.cs
```diff
@@ -106,6 +107,25 @@ public Task StopAsync(CancellationToken cancellationToken)
     AddGauge("",
              QueuedName);
 
+    // initialize all the gauges
```
ngruelaneo, line 110:
```suggestion
    // initialize all the gauges for each partition
```

ngruelaneo (approved):

PR #141, opened 2022-09-02
Better management of sigterm

aneojgurhem:
fix https://github.com/aneoconsulting/ArmoniK/issues/467


Common/src/Pollster/Pollster.cs
```diff
@@ -151,6 +151,11 @@ await foreach (var message in messages.WithCancellation(cancellationToken)
 
           var combinedCts = CancellationTokenSource.CreateLinkedTokenSource(message.CancellationToken,
                                                                             cancellationToken);
+          if (combinedCts.IsCancellationRequested)
+          {
+            return;
```
lemaitre-aneo, line 156: Do we really want a return? or throw a `OperationCanceledException`?

aneojgurhem: Yes, at this point we do not want to acquire the task. So we quit. Maybe break would have been more appropriate. and put the message at postponed

aneojgurhem: well, break break because there is a foreach and a while

aneojgurhem: I really don't like the break; break;

Adaptors/MongoDB/src/TaskTable.cs
```diff
@@ -588,9 +588,9 @@ await foreach (var taskId in taskCollection.AsQueryable(sessionHandle)
                                                                                                           TaskStatus.Dispatched,
                                                                                                         }));
 
-    Logger.LogDebug("Release task {task} on {podName}",
-                    taskId,
-                    ownerPodId);
+    Logger.LogInformation("Release task {task} on {podName}",
```
wkirschenmann, line 591: Isn't `task` already in the log scope ?
BTW, maybe it'd be a good Idea to have `podname` in a wider scope.

Common/src/Pollster/Pollster.cs
```diff
@@ -199,12 +202,27 @@ await taskHandler.ReleaseTask(CancellationToken.None)
                 return;
               }
 
-              await taskHandler.ExecuteTask(combinedCts.Token)
+              var workerConnectionCts = new CancellationTokenSource();
+              var postConnectionCts = new CancellationTokenSource();
+              var errorConnectionCts = new CancellationTokenSource();
+              combinedCts.Token.Register(() =>
```
wkirschenmann, line 208: You should register for the CancellationToken of main loop instead. This one includes a cancellationToken from the message (is it still relevant anyway with AMQP ?).

Common/src/Pollster/TaskHandler.cs
```diff
@@ -408,14 +408,14 @@ public async Task PreProcessing(CancellationToken cancellationToken)
...
-  /// <param name="cancellationToken">Token used to cancel the execution of the method</param>
+  /// <param name="cancellationToken">Token used to cancel the method before the task is started</param>
+  /// <param name="connectionCancellationToken">Token used to cancel the connection to the worker</param>
   /// <returns>
   ///   Task representing the asynchronous execution of the method
   /// </returns>
   /// <exception cref="NullReferenceException">wrong order of execution</exception>
   /// <exception cref="ArmoniKException">worker pipe is not initialized</exception>
-  public async Task ExecuteTask(CancellationToken cancellationToken)
+  public async Task ExecuteTask(CancellationToken cancellationToken, CancellationToken connectionCancellationToken)
```
wkirschenmann, line 418: Why not have the grace delay defined here instead of passing 2 different cancellationTokens ?

Common/src/Pollster/TaskHandler.cs
```diff
@@ -463,31 +463,44 @@ await workerStreamHandler_.Pipe.CompleteAsync()
...
+  /// <param name="postCancellationToken">Token used to cancel the operations after the reply is received</param>
+  /// <param name="errorCancellationToken">Token used to cancel the error management operations</param>
   /// <returns>
   ///   Task representing the asynchronous execution of the method
   /// </returns>
   /// <exception cref="NullReferenceException">wrong order of execution</exception>
-  public async Task PostProcessing(CancellationToken cancellationToken)
+  public async Task PostProcessing(CancellationToken readCancellationToken,
+                                   CancellationToken postCancellationToken,
+                                   CancellationToken errorCancellationToken)
```
wkirschenmann, line 475: Same question ?

Common/src/Pollster/TaskHandler.cs
wkirschenmann: I would set the two cancellationtokens (from the message and from sigterm) directly into the constructor of this class. That way you can have all the logic into one sigle class.

Common/src/Pollster/TaskHandler.cs
```diff
@@ -463,31 +463,44 @@ await workerStreamHandler_.Pipe.CompleteAsync()
...
   {
     using var _ = logger_.BeginNamedScope("PostProcessing",
                                           ("taskId", messageHandler_.TaskId));
 
-    if (workerStreamHandler_.Pipe == null || taskData_ == null || agent_ == null)
+    if (workerStreamHandler_.Pipe is null)
```
wkirschenmann, line 480: I am not sure about the relevance of these tests. They should have been done before (ex: line 444).
However, we should check that before postprocessing, the processing has been finished.

Common/src/Pollster/TaskHandler.cs
```diff
@@ -496,14 +509,14 @@ await agentHandler_.Stop(CancellationToken.None)
...
-        await agent_.FinalizeTaskCreation(CancellationToken.None)
+        await agent_.FinalizeTaskCreation(postCancellationToken)
                     .ConfigureAwait(false);
       }
 
       await submitter_.CompleteTaskAsync(taskData_,
                                          false,
                                          reply.Output,
-                                         CancellationToken.None)
+                                         postCancellationToken)
```
wkirschenmann, line 519: In these last calls, I'd rather have no cancellationtoken : these are really the very last thing that are being done for the task. If there is a chance, even small that it can be done before the container is killed, we should take it.

Common/tests/Helpers/TestPollsterProvider.cs
```diff
@@ -0,0 +1,164 @@
...
+
+namespace ArmoniK.Core.Common.Tests.Helpers;
+
+public class TestPollsterProvider : IDisposable
+{
+  private const           string                   DatabaseName   = "ArmoniK_TestDB";
+  private static readonly ActivitySource           ActivitySource = new("ArmoniK.Core.Common.Tests.TestPollsterProvider");
+  private readonly        WebApplication           app_;
+  private readonly        IMongoClient             client_;
+  private readonly        ILoggerFactory           loggerFactory_;
```
wkirschenmann, line 59: Lots of github actions warnings bellow

Adaptors/MongoDB/src/TaskTable.cs
aneojgurhem: TaskId should be in the scope. I added the podId in the scope. But further rework on logs is out of scope. We need to improve our logs but it is a issue in itself.

Common/tests/Helpers/TestPollsterProvider.cs
aneojgurhem: Some are bugged, some are fixed in another pr :
image

Common/src/Pollster/TaskHandler.cs
aneojgurhem: We are checking in every method due to nullable warnings.  I changed it to this format because I had one of them null while passing mocks but did not know which one was missing.

Common/src/Pollster/TaskHandler.cs
aneojgurhem: In here, I am only using the combined one and the one to manage the grace delay

Common/src/Pollster/TaskHandler.cs
wkirschenmann: OK. Keep it that way untill a full refactor of the TaskHandler : acquisition and processing should be split.

Compute/PollingAgent/src/Program.cs
```diff
@@ -48,6 +48,7 @@
 
 using Serilog;
 
+using Pollster = ArmoniK.Core.Common.Injection.Options.Pollster;
```
wkirschenmann, line 51: Not sure that options should take precedence over the concrete types they apply to.

aneojgurhem: I can't. Resharper is configured like so. The alternative is more verbose so I think r# chooses this one

Common/src/Injection/Options/Pollster.cs
```diff
@@ -0,0 +1,34 @@
...
+// GNU Affero General Public License for more details.
+// 
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+using System;
+
+namespace ArmoniK.Core.Common.Injection.Options;
+
+public class Pollster
```
ngruelaneo, line 29: Missing documentation

Common/src/Pollster/Pollster.cs
ngruelaneo: return est pas si mal du coup

(ps: rigolo la meme discussion a eu lieu sur un autre langage que wil "adore" :)

PS:  on a ete bien conditionné à éviter l'utilisation des goto...

Common/tests/Helpers/ConsoleForwardingLoggerProvider.cs
```diff
@@ -41,7 +41,7 @@ public ConsoleForwardingLoggerProvider(LogLevel minLogLevel)
                                                 {
                                                   if (logLevel >= minLogLevel)
                                                   {
-                                                    Console.WriteLine(logLevel + " => " + category + "\n" + message + "\n" + exception);
+                                                    Console.WriteLine(logLevel + " : " + DateTime.Now + " => " + category + "\n" + message + "\n" + exception);
```
lemaitre-aneo, line 44: format string?

lemaitre-aneo (approved):

PR #147, opened 2022-09-15
Add result service

aneojgurhem:
fix https://github.com/aneoconsulting/ArmoniK/issues/498

### DIFF
--- a/Adaptors/MongoDB/src/ResultTable.cs
+++ b/Adaptors/MongoDB/src/ResultTable.cs
@@ -94,9 +94,9 @@ public class ResultTable : IResultTable
     }
   }
 
-  public async IAsyncEnumerable<Core.Common.Storage.Result> GetResults(string                                     sessionId,
-                                                                       IEnumerable<string>                        keys,
-                                                                       [EnumeratorCancellation] CancellationToken cancellationToken = default)
+  public IAsyncEnumerable<Core.Common.Storage.Result> GetResults(string              sessionId,
+                                                                 IEnumerable<string> keys,
+                                                                 CancellationToken   cancellationToken = default)
   {
     using var activity = activitySource_.StartActivity($"{nameof(GetResults)}");
     activity?.SetTag($"{nameof(GetResults)}_sessionId",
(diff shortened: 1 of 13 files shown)


Adaptors/MongoDB/src/ResultTable.cs
```diff
@@ -94,30 +94,29 @@ public class ResultTable : IResultTable
...
-                                   .SingleAsync(cancellationToken)
-                                   .ConfigureAwait(false);
-    }
-    catch (InvalidOperationException)
+    var cursor = await resultCollection.AsQueryable(sessionHandle)
+                                       .Where(model => keys.Contains(model.Name) && model.SessionId == sessionId)
+                                       .ToCursorAsync(cancellationToken)
+                                       .ConfigureAwait(false);
+
+    while (await cursor.MoveNextAsync(cancellationToken)
```
lemaitre-aneo, line 112: Could this be an extension method of a cursor? I expect this to be used many times.

Common/src/Storage/IResultTable.cs
```diff
@@ -195,3 +198,38 @@ public interface IResultTable : IInitializable
...
+  ///   Get the result from its id
+  /// </summary>
+  /// <param name="resultTable"></param>
+  /// <param name="sessionId">id of the session containing the result</param>
+  /// <param name="key">id of the result to be retrieved</param>
+  /// <param name="cancellationToken">Token used to cancel the execution of the method</param>
+  /// <returns>
+  ///   Result metadata from the database
+  /// </returns>
+  public static async Task<Result> GetResult(this IResultTable resultTable,
```
lemaitre-aneo, line 214: I think it is better to put this in the interface. If an adapter can be optimized in this case, it will be able to override it.

Common/src/gRPC/Services/GrpcResultsService.cs
```diff
@@ -0,0 +1,76 @@
...
+  private readonly IResultTable                resultTable_;
+
+  public GrpcResultsService(IResultTable                resultTable,
+                            ILogger<GrpcResultsService> logger)
+  {
+    logger_      = logger;
+    resultTable_ = resultTable;
+  }
+
+  public override async Task<GetOwnerTaskIdResponse> GetOwnerTaskId(GetOwnerTaskIdRequest request,
```
lemaitre-aneo, line 53: I think the API should be `getResult` directly, which returns all the info about the result, not just its `ownerTaskId`.

lemaitre-aneo (changes requested):

Common/src/gRPC/Services/GrpcResultsService.cs
aneojgurhem: This was to reduce the data in transit

Common/src/Storage/IResultTable.cs
aneojgurhem: I wanted to reduce code duplication but ok

lemaitre-aneo: There is no code duplication if you put it in the interface (same file).

Common/src/gRPC/Services/GrpcResultsService.cs
lemaitre-aneo: A large a message would be? I believe less than a kilobyte, so less than a tcp packet. This is fine.

Common/src/Storage/IResultTable.cs
aneojgurhem: this is the interface file

Common/src/gRPC/Services/GrpcResultsService.cs
aneojgurhem: It returns multiple values

lemaitre-aneo: my bad.

lemaitre-aneo (approved):

lemaitre-aneo (approved):

lemaitre-aneo (approved):

PR #148, opened 2022-09-20
avoid dereferencing null object

jfonseca-aneo:
Fix https://github.com/aneoconsulting/ArmoniK/issues/513

### DIFF
--- a/Control/PartitionMetrics/src/ArmoniKMeter.cs
+++ b/Control/PartitionMetrics/src/ArmoniKMeter.cs
@@ -127,7 +127,11 @@ public class ArmoniKMeter : Meter, IHostedService
     await foreach (var partition in partitionTable_.GetPartitionWithAllocationAsync(cancellationToken)
                                                    .ConfigureAwait(false))
     {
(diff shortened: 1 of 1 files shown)


aneojgurhem (dismissed):

Control/PartitionMetrics/src/ArmoniKMeter.cs
```diff
@@ -127,7 +127,11 @@ public Task StopAsync(CancellationToken cancellationToken)
     await foreach (var partition in partitionTable_.GetPartitionWithAllocationAsync(cancellationToken)
                                                    .ConfigureAwait(false))
     {
-      var metricValue = (metrics[$"armonik_{partition.PartitionId}_tasks_queued"] as Metric)!.MetricValue;
+      var metric = metrics[$"armonik_{partition.PartitionId}_tasks_queued"] as Metric;
```
lemaitre-aneo, line 130: You can just use the `?` operator.
```suggestion
      var metricValue = (metrics[$"armonik_{partition.PartitionId}_tasks_queued"] as Metric)?.MetricValue;
```

jfonseca-aneo: yes! that was indeed ugly

wkirschenmann (approved):

PR #149, opened 2022-09-20
Add service for getting expected output keys

aneojgurhem:


Common/src/Storage/ITaskTable.cs
```diff
@@ -300,8 +312,25 @@ public interface ITaskTable : IInitializable
...
+  {
+    try
+    {
+      return await GetTasksExpectedOutputKeys(new[]
+                                              {
+                                                taskId,
+                                              },
+                                              cancellationToken)
+                   .Select(tuple => tuple.expectedOutputKeys)
+                   .SingleAsync(cancellationToken)
```
lemaitre-aneo, line 326: `SingleAsync` throws if sequence is empty, but Excpection would not give the right information (ie: missing task ID).

aneojgurhem: There is a catch for this exception !

Common/src/gRPC/Services/GrpcTasksService.cs
```diff
@@ -121,4 +120,44 @@ public class GrpcTasksService : Tasks.TasksBase
...
+               {
+                 await taskTable_.GetTasksExpectedOutputKeys(request.TaskId,
+                                                             context.CancellationToken)
+                                 .Select(r => new GetResultIdsResponse.Types.MapTaskResult
+                                              {
+                                                ResultIds =
+                                                {
+                                                  r.expectedOutputKeys,
+                                                },
+                                                TaskId = r.taskId,
```
wkirschenmann, line 141: Just for coherence with the object name, I'd put the TaskId before the ResultIds

Adaptors/MongoDB/src/TaskTable.cs
```diff
@@ -626,6 +624,24 @@ await foreach (var taskId in taskCollection.AsQueryable(sessionHandle)
...
 
+  public IAsyncEnumerable<(string taskId, IEnumerable<string> expectedOutputKeys)> GetTasksExpectedOutputKeys(IEnumerable<string> taskIds,
+                                                                                                              CancellationToken   cancellationToken = default)
+  {
+    using var activity       = activitySource_.StartActivity($"{nameof(GetTasksExpectedOutputKeys)}");
+    var       sessionHandle  = sessionProvider_.Get();
+    var       taskCollection = taskCollectionProvider_.Get();
+
+    return taskCollection.AsQueryable(sessionHandle)
+                         .Where(tdm => taskIds.Contains(tdm.TaskId))
```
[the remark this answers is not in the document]
lemaitre-aneo: The question was: if we have a `IMongoQueryable`, are we sure that the function used is the one from mongo, and not the one from IAsyncEnumerable. I'm fine to defer this discussion.

lemaitre-aneo (dismissed):

lemaitre-aneo (approved):

wkirschenmann (approved):

PR #150, opened 2022-09-20
RabittMQ adapter

jfonseca-aneo:


Adaptors/RabbitMQ/tests/QueueStorageTests.cs
```diff
@@ -0,0 +1,255 @@
...
+using Microsoft.Extensions.Diagnostics.HealthChecks;
+using Microsoft.Extensions.Logging.Abstractions;
+
+using Moq;
+
+using NUnit.Framework;
+
+using RabbitMQ.Client;
+
+namespace ArmoniK.Core.Adapters.RabbitMQ.Tests;
```
aneojgurhem, line 41: Ces tests ont l'air d'être un copie de ceux d'AMQP. Est ce qu'on peut les mutualiser ?

Adaptors/Amqp/src/QueueMessageHandler.cs
```diff
@@ -59,7 +59,7 @@ public class QueueMessageHandler : IQueueMessageHandler
   }
 
   /// <inheritdoc />
-  public CancellationToken CancellationToken { get; }
+  public CancellationToken CancellationToken { get; set; }
```
aneojgurhem, line 62: Why set ?

Adaptors/RabbitMQ/src/PullQueueStorage.cs
```diff
@@ -0,0 +1,197 @@
...
+  {
+    var nbPulledMessage = 0;
+
+    if (!IsInitialized)
+    {
+      throw new ArmoniKException($"{nameof(PullQueueStorage)} should be initialized before calling this method.");
+    }
+
+    // TODO: Fix this, I put it here cuz the method must be async to return an IAsyncEnumerable.
+    await Task.CompletedTask.ConfigureAwait(false);
```
aneojgurhem, line 180: Maybe try ValueTask that are lighter than Task. Or juste suppress the warning

Adaptors/RabbitMQ/src/PushQueueStorage.cs
```diff
@@ -0,0 +1,91 @@
...
+
+    ConnectionRabbit.Channel!.ExchangeDeclare("ArmoniK.QueueExchange",
+                                              "direct");
+
+    var basicProperties = ConnectionRabbit.Channel!.CreateBasicProperties();
+    basicProperties.Priority = Convert.ToByte(priority);
+
+    foreach (var msg in messages)
+    {
+      ConnectionRabbit.Channel.BasicPublish("ArmoniK.QueueExchange",
```
aneojgurhem, line 85: No batch method ?

Adaptors/RabbitMQ/src/QueueMessageHandler.cs
```diff
@@ -0,0 +1,104 @@
...
+      case QueueMessageStatus.Postponed:
+        /* Negative acknowledging this message will send it
+         to the retry exchange, see PullQueueStorage.cs */
+        channel_.BasicNack(deliverEvent_.DeliveryTag,
+                           false,
+                           false);
+        break;
+      case QueueMessageStatus.Failed:
+
+      case QueueMessageStatus.Processed:
```
aneojgurhem, line 86: I would suggest a rework and documentation of those statuses

docker-compose/docker-compose.override.yml
```diff
@@ -4,32 +4,46 @@ services:
...
+      - Components__QueueStorage=ArmoniK.Adapters.Amqp.QueueStorage
 
   armonik.control.metrics:
     environment:
       - ASPNETCORE_ENVIRONMENT=Development
 
   armonik.compute.pollingagent0:
+    depends_on:
+      queue:
+        condition: service_started
```
aneojgurhem, line 16: Why in the override and not the regular docker-compose.yml ?
A bit of explanation for this should be better. Does it uses our health checks ?

aneojgurhem (changes requested):

Adaptors/Amqp/src/QueueMessageHandler.cs
aneojgurhem: I don't remember the usage of the set part.

docker-compose/docker-compose.override.yml
jfonseca-aneo: It is already in the regular compose file. I forgot to remove it from here ^^. I'll add some doc.

Adaptors/RabbitMQ/src/PushQueueStorage.cs
jfonseca-aneo: The .net client does not provide with an implementation, but in their git repository they show how to do it in [this example](https://github.com/rabbitmq/rabbitmq-dotnet-client/blob/main/projects/TestApplications/MassPublish/Program.cs). We can try that out, but in a separate PR I'd say

Adaptors/RabbitMQ/src/PullQueueStorage.cs
jfonseca-aneo: Ok; I'll give it a try

Adaptors/RabbitMQ/src/PushQueueStorage.cs
aneojgurhem: ok

Adaptors/Amqp/src/QueueMessageHandler.cs
jfonseca-aneo: I put the set to pass the right cancellation token in the RabbitMQ adapter, its only used there.

Adaptors/RabbitMQ/src/QueueMessageHandler.cs
jfonseca-aneo: Agree... but in another PR

aneojgurhem (approved):

PR #151, opened 2022-09-21
Specialize mongo getresult

lemaitre-aneo:
The factorization of GetResult and GetResults resulted in a very big slow down of Extension Csharp.
This PR specializes GetResult for MongoDB in order to have a faster DB request.

GetResult 17 times faster

### DIFF
--- a/Adaptors/MongoDB/src/ResultTable.cs
+++ b/Adaptors/MongoDB/src/ResultTable.cs
@@ -94,6 +94,31 @@ public class ResultTable : IResultTable
     }
   }
 
+  async Task<Core.Common.Storage.Result> IResultTable.GetResult(string              sessionId,
+                                                                 string key,
+                                                                 CancellationToken   cancellationToken)
+  {
+    using var activity = activitySource_.StartActivity($"{nameof(IResultTable.GetResult)}");
+    activity?.SetTag($"{nameof(IResultTable.GetResult)}_sessionId",
+                     sessionId);
+    var sessionHandle    = sessionProvider_.Get();
+    var resultCollection = resultCollectionProvider_.Get();
+    try
+    {
+      return await resultCollection.AsQueryable(sessionHandle)
+                                   .Where(model => model.Id == Result.GenerateId(sessionId,
+                                                                                 key))
+                                   .SingleAsync(cancellationToken)
+                                   .ConfigureAwait(false);
+    }
+    catch (InvalidOperationException)
+    {
+      throw new ResultNotFoundException($"Key '{key}' not found");
+    }
+  }
+  
+
+
   public IAsyncEnumerable<Core.Common.Storage.Result> GetResults(string              sessionId,
                                                                  IEnumerable<string> keys,
                                                                  CancellationToken   cancellationToken = default)
--- a/Adaptors/MongoDB/src/ResultTable.cs
+++ b/Adaptors/MongoDB/src/ResultTable.cs
@@ -94,9 +94,9 @@ public class ResultTable : IResultTable
     }
   }
 
-  async Task<Core.Common.Storage.Result> IResultTable.GetResult(string              sessionId,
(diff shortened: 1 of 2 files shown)


aneojgurhem (approved):

Adaptors/MongoDB/src/ResultTable.cs
```diff
@@ -94,6 +94,30 @@ public class ResultTable : IResultTable
     }
   }
 
+  async Task<Core.Common.Storage.Result> IResultTable.GetResult(string            sessionId,
+                                                                string            key,
+                                                                CancellationToken cancellationToken)
```
wkirschenmann, line 99: =default

Adaptors/MongoDB/src/ResultTable.cs
```diff
@@ -94,6 +94,30 @@ public class ResultTable : IResultTable
...
+  {
+    using var activity = activitySource_.StartActivity($"{nameof(IResultTable.GetResult)}");
```
wkirschenmann, line 101: nameof(GetResult)

Adaptors/MongoDB/src/ResultTable.cs
```diff
@@ -94,6 +94,30 @@ public class ResultTable : IResultTable
...
+    activity?.SetTag($"{nameof(IResultTable.GetResult)}_sessionId",
```
wkirschenmann, line 102: nameof(GetResult)

Adaptors/MongoDB/src/ResultTable.cs
```diff
@@ -94,6 +94,30 @@ public class ResultTable : IResultTable
...
+                     sessionId);
+    var sessionHandle    = sessionProvider_.Get();
+    var resultCollection = resultCollectionProvider_.Get();
+    try
+    {
+      return await resultCollection.AsQueryable(sessionHandle)
+                                   .Where(model => model.Id == Result.GenerateId(sessionId,
+                                                                                 key))
+                                   .SingleAsync(cancellationToken)
```
wkirschenmann, line 111: This request is very different than the one from GetResults (use `model.Id`). Does using this approach in the `GetResults` approach lead to better performances (better usage of indexes) ?

Adaptors/MongoDB/src/ResultTable.cs
aneojgurhem: According to R#, not necessary here !

Adaptors/MongoDB/src/ResultTable.cs
aneojgurhem: Not working

Adaptors/MongoDB/src/ResultTable.cs
aneojgurhem: idem

wkirschenmann: Ok. Strange but OK.

Adaptors/MongoDB/src/ResultTable.cs
wkirschenmann: I understand why R# says that it's not necessary. But I'm not sure it's true.
[https://stackoverflow.com/a/8909925](https://stackoverflow.com/a/8909925)

lemaitre-aneo: Optional parameters are given their default value by the caller, not the callee. And the caller does not know what is the actual type of the object, and thus, can only take the default value specified in the base declaration.

Adaptors/MongoDB/src/ResultTable.cs
lemaitre-aneo: We can have a look later. But the performance of the batched version seems good, so I'm not worried about that. The goal of this PR was to fix a regression.

PR #152, opened 2022-09-26
Rework health checks

aneojgurhem:
fix https://github.com/aneoconsulting/ArmoniK/issues/431

This PR fixes the error shown in the issue above. It also modifies how we define health checks (HC) to have a better transmission of the information associated to the success or failure of the check. (fix https://github.com/aneoconsulting/ArmoniK/issues/188)

Before this PR, the HC was checking only the initialization of the MongoDB session provider in control plane and polling agent.
Now, it also checks the connection between the worker and the polling agent. It also checks the connection to MongoDB and Redis.
The polling agent now has a health check. This health check is used to stop task acquisition when health check fails.

It also adds tests for HC (fix https://github.com/aneoconsulting/ArmoniK/issues/160)


Adaptors/Amqp/src/PullQueueStorage.cs
```diff
@@ -90,8 +91,10 @@ public class PullQueueStorage : QueueStorage, IPullQueueStorage
     logger_ = logger;
   }
 
-  public new ValueTask<bool> Check(HealthCheckTag tag)
-    => ValueTask.FromResult(isInitialized_);
+  public new Task<HealthCheckResult> Check(HealthCheckTag tag)
```
lemaitre-aneo, line 94: What does this `new` means?

Adaptors/Amqp/src/PullQueueStorage.cs
```diff
@@ -90,8 +91,10 @@ public class PullQueueStorage : QueueStorage, IPullQueueStorage
...
+    => Task.FromResult(isInitialized_
```
lemaitre-aneo, line 95: I would expect a test a bit more involved here.

aneojgurhem: I took the initial one. Most of them are like that. Moreover this health check is not registered.

Adaptors/Amqp/src/PullQueueStorage.cs
aneojgurhem: It seems to say that you want to implement another version of this method

aneojgurhem: https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/knowing-when-to-use-override-and-new-keywords

Adaptors/Amqp/src/PullQueueStorage.cs
aneojgurhem: I will not modify this adaptor too much because of https://github.com/aneoconsulting/ArmoniK.Core/pull/150

Common/src/HealthCheck.cs
```diff
@@ -51,17 +51,24 @@ public class HealthCheck : IHealthCheck
...
 
     return context.Registration.FailureStatus switch
            {
-             HealthStatus.Unhealthy => HealthCheckResult.Unhealthy(),
-             HealthStatus.Degraded  => HealthCheckResult.Degraded(),
-             HealthStatus.Healthy   => HealthCheckResult.Healthy(),
+             HealthStatus.Unhealthy => HealthCheckResult.Unhealthy(result.Description,
+                                                                   result.Exception,
+                                                                   result.Data),
+             HealthStatus.Degraded => HealthCheckResult.Degraded(result.Description,
```
ngruelaneo, line 67: cela ne devrait jamais se passer non ? (cf ligne 57)

aneojgurhem: ça se passe. J'ai un test qui passe dessus. Il y a Unhealthy dans le HealthStatus

ngruelaneo (approved):

Common/tests/Auth/AuthenticationTableTestBase.cs
```diff
@@ -192,6 +195,36 @@ public virtual void GetAuthSource()
   {
   }
 
+  [Test]
+  public async Task InitShouldSucceed()
+  {
+    if (RunTests)
+    {
+      Assert.AreEqual(HealthStatus.Unhealthy,
```
lemaitre-aneo, line 203: Maybe `Assert.AreNotEqual(HealthStatus.Healthy`?

lemaitre-aneo (approved):

PR #154, opened 2022-09-29
Add missing doc

aneojgurhem:
Improves the documentation of public items. It reduces the number of warnings that comes from the addition of `<GenerateDocumentationFile>true</GenerateDocumentationFile>` in the `.csproj` file of the Common project. One of our goals is to eliminate all the warning due to this option by populating all the missing documentation fields.

fix https://github.com/aneoconsulting/ArmoniK/issues/577

### DIFF
--- a/Common/src/Injection/ServiceCollectionExt.cs
+++ b/Common/src/Injection/ServiceCollectionExt.cs
@@ -40,13 +40,36 @@ using Microsoft.Extensions.Diagnostics.HealthChecks;
 
 namespace ArmoniK.Core.Common.Injection;
 
+/// <summary>
+///   Extends the functionality of the <see cref="IConfiguration" />
+/// </summary>
 public static class ConfigurationExt
 {
+  /// <summary>
+  ///   Configure an object with the given configuration.
+  /// </summary>
+  /// <typeparam name="T">Type of the options class</typeparam>
+  /// <param name="configuration">Configurations used to fill the class</param>
+  /// <param name="key">Path to the Object in the configuration</param>
+  /// <returns>
+  ///   The initialized object
+  /// </returns>
+  /// <exception cref="InvalidOperationException">the <paramref name="key" /> is not found in the configurations.</exception>
   public static T GetRequiredValue<T>(this IConfiguration configuration,
                                       string              key)
     => configuration.GetRequiredSection(key)
                     .Get<T>();
 
+  /// <summary>
+  ///   Configure an object with the given configuration.
+  ///   If the object is not found in the configuration, a new object in returned.
+  /// </summary>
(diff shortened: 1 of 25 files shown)


Common/src/Injection/ServiceCollectionExt.cs
```diff
@@ -40,20 +40,46 @@
...
+/// <summary>
+///   Extends the functionality of the <see cref="IConfiguration" />
+/// </summary>
 public static class ConfigurationExt
 {
+  /// <summary>
+  ///   Configure an object with the given configuration.
+  /// </summary>
+  /// <typeparam name="T">Type of the options class</typeparam>
+  /// <param name="configuration">Configurations used to fill the class</param>
```
jfonseca-aneo, line 52: Configuration properties used to populate the class

Common/src/Injection/ServiceCollectionExt.cs
```diff
@@ -40,20 +40,46 @@
...
                                       string              key)
     => configuration.GetRequiredSection(key)
                     .Get<T>();
 
+  /// <summary>
+  ///   Configure an object with the given configuration.
+  ///   If the object is not found in the configuration, a new object in returned.
+  /// </summary>
+  /// <typeparam name="T">Type of the options class</typeparam>
+  /// <param name="configuration">Configurations used to fill the class</param>
```
jfonseca-aneo, line 68: idem

Common/src/Injection/ServiceCollectionExt.cs
```diff
@@ -116,6 +170,15 @@ public static class ServiceCollectionExt
     return services;
   }
 
+  /// <summary>
+  ///   Add a singleton service of the type specified with health check capabilities
```
jfonseca-aneo, line 174: of the specified type

jfonseca-aneo: idem the next 4 references

Common/src/Utils/LocalIPv4.cs
```diff
@@ -57,6 +71,12 @@ public static string GetLocalIPv4(NetworkInterfaceType _type)
     return output;
   }
 
+  /// <summary>
+  ///   Get local IP from Ethernet network interface.
```
jfonseca-aneo, line 75: Should we add that we identify the owner pod with this value?

aneojgurhem: Not in this class. This is a general usage class that may not be linked to our usage of it.

jfonseca-aneo: Ok

jfonseca-aneo (dismissed):

PR #157, opened 2022-10-03
Rework Amqp adapter to expose its health check and use health check to check service statuses after stopping them

aneojgurhem:
The purpose of this pr is to make sure that the health checks for the control plane and the polling agent fail when the queue/database/object storage are down. It also adds a pipeline test in docker-compose to call the health check in a more complete environment compared to the Unit Tests.

fix https://github.com/aneoconsulting/ArmoniK/issues/579


Adaptors/Amqp/src/ConnectionAmqp.cs
```diff
@@ -77,9 +103,46 @@ public async Task<IConnectionAmqp> OpenConnectionAsync()
...
+        Connection.AddClosedCallback((x,
+                                      e) => OnCloseConnection(x,
+                                                              e,
+                                                              logger_));
+        break;
+      }
+      catch
+      {
+        logger_.LogInformation("Retrying to create connection");
+        Thread.Sleep(1000 * retry);
```
lemaitre-aneo, line 122:
```suggestion
        Task.Delay(TimeSpan.FromSeconds(retry))
```
`Thread.Sleep` is blocking.

Adaptors/Amqp/src/PullQueueStorage.cs
```diff
@@ -46,66 +45,64 @@ public class PullQueueStorage : QueueStorage, IPullQueueStorage
...
+                             .ToArray();
+
+      senders_ = Enumerable.Range(0,
+                                  NbLinks)
+                           .Select(i => new AsyncLazy<ISenderLink>(() => new SenderLink(session,
+                                                                                        $"{Options.PartitionId}###SenderLink{i}",
+                                                                                        $"{Options.PartitionId}###q{i}")))
+                           .ToArray();
+
       var senders   = Task.WhenAll(senders_.Select(async lazy => await lazy));
```
lemaitre-aneo, line 100: Couldn't you just do `Task.WhenAll(senders_)` ?

Adaptors/Amqp/src/PullQueueStorage.cs
```diff
@@ -116,10 +113,15 @@ public override async Task Init(CancellationToken cancellationToken)
...
+    }
+
     while (nbPulledMessage < nbMessages)
     {
       var currentNbMessages = nbPulledMessage;
-      for (var i = receivers_.Length - 1; i >= 0; --i)
+      for (var i = receivers_!.Length - 1; i >= 0; --i)
       {
         cancellationToken.ThrowIfCancellationRequested();
         var receiver = await receivers_[i];
```
lemaitre-aneo, line 127: As `Init` is always called at this point, maybe `receivers_` and `senders_` could be directly a list `ReceiverLink` and `SenderLink` instead of tasks. This would avoid the `await` here.

Adaptors/Amqp/src/PullQueueStorage.cs
```diff
@@ -46,66 +45,64 @@ public class PullQueueStorage : QueueStorage, IPullQueueStorage
 {
   private readonly ILogger<PullQueueStorage> logger_;
 
-  private readonly AsyncLazy<IReceiverLink>[] receivers_;
-  private readonly AsyncLazy<ISenderLink>[]   senders_;
-
-  private bool isInitialized_;
+  private AsyncLazy<IReceiverLink>[]? receivers_;
```
lemaitre-aneo, line 48: Do you really need to have an optional? Isn't an empty array not enough?

Adaptors/Amqp/src/SessionAmqp.cs
```diff
@@ -1,164 +0,0 @@
-// This file is part of the ArmoniK project
```
lemaitre-aneo, line 1: Where the functionality of this file went?

Adaptors/Amqp/src/ConnectionAmqp.cs
aneojgurhem: Needs an await though

Adaptors/Amqp/src/PullQueueStorage.cs
aneojgurhem: No there is the alzy async thingy in it

lemaitre-aneo: I thought `AsyncLazy` was awaitable.

Adaptors/Amqp/src/SessionAmqp.cs
aneojgurhem: new Session(connection); No need for that complexity

lemaitre-aneo (approved):

PR #162, opened 2022-10-07
Add more doc

aneojgurhem:
Improves the documentation of public items. It reduces the number of warnings that comes from the addition of `<GenerateDocumentationFile>true</GenerateDocumentationFile>` in the `.csproj` file of the Common project. One of our goals is to eliminate all the warning due to this option by populating all the missing documentation fields.

fix https://github.com/aneoconsulting/ArmoniK/issues/577

### DIFF
--- a/Common/src/Storage/IPullQueueStorage.cs
+++ b/Common/src/Storage/IPullQueueStorage.cs
@@ -38,7 +38,7 @@ public interface IPullQueueStorage : IQueueStorage
   /// <param name="nbMessages">Number of messages to retrieve</param>
   /// <param name="cancellationToken">Token used to cancel the execution of the method</param>
   /// <returns>
-  ///   Async Enumerable of queue messages
+  ///   Enumerator allowing async iteration over the message queue
   /// </returns>
   IAsyncEnumerable<IQueueMessageHandler> PullMessagesAsync(int               nbMessages,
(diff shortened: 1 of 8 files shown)


Common/src/Storage/IPullQueueStorage.cs
```diff
@@ -27,8 +27,19 @@
...
+/// </summary>
 public interface IPullQueueStorage : IQueueStorage
 {
+  /// <summary>
+  ///   Gets messages from the queue
+  /// </summary>
+  /// <param name="nbMessages">Number of messages to retrieve</param>
+  /// <param name="cancellationToken">Token used to cancel the execution of the method</param>
+  /// <returns>
+  ///   Async Enumerable of queue messages
```
jfonseca-aneo, line 41: Enumerator allowing async iteration over the message queue

Common/src/Storage/Output.cs
```diff
@@ -27,9 +27,18 @@
 
 namespace ArmoniK.Core.Common.Storage;
 
+/// <summary>
+///   Represents the output of a task
```
jfonseca-aneo, line 31: I'd suggest: Record encoding if a task successfully produced an output

jfonseca-aneo (approved):

PR #170, opened 2022-10-25
Mark the pollster as failed when worker is unavailable

aneojgurhem:
When the worker is down, the polling agent was still trying to send tasks to it. Then the tasks failed and the polling agent was putting them in the error status. It was not detecting that there was an issue with the worker, therefore putting multiple tasks in error whereas they would have been able to succeed.
This PRs marks the polling agent as unhealthy and stop sending tasks to the worker when it detects that the worker is unavailable.

fix https://github.com/aneoconsulting/ArmoniK/issues/636

### DIFF
--- a/Common/src/Pollster/Pollster.cs
+++ b/Common/src/Pollster/Pollster.cs
@@ -320,7 +320,6 @@ public class Pollster : IInitializable
             catch (RpcException e) when (e.StatusCode == StatusCode.Unavailable)
             {
               // This exception should stop pollster
-              RecordError(e);
               throw;
             }
             catch (Exception e)
@@ -336,6 +335,9 @@ public class Pollster : IInitializable
         catch (RpcException e) when (e.StatusCode == StatusCode.Unavailable)
         {
           // This exception should stop pollster
+          healthCheckFailedResult_ = HealthCheckResult.Unhealthy("Worker unavailable",
+                                                                 e);
+          cts.Cancel();
           throw;
         }
         catch (TooManyException)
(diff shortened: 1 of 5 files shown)


jfonseca-aneo (dismissed):

Common/src/Pollster/Pollster.cs
```diff
@@ -315,6 +317,12 @@ await taskHandler.PostProcessing()
                 }
               }
             }
+            catch (RpcException e) when (e.StatusCode == StatusCode.Unavailable)
+            {
+              // This exception should stop pollster
+              RecordError(e);
```
lemaitre-aneo, line 323:
```suggestion
      healthCheckFailedResult_ = HealthCheckResult.Unhealthy("...");
      cts.Cancel();
```

Common/tests/Pollster/PollsterTest.cs
```diff
@@ -568,4 +575,63 @@ await pollster.Init(CancellationToken.None)
     Assert.AreSame(string.Empty,
                    pollster.TaskProcessing);
   }
+
+
+  [Test]
+  public async Task UnavailableWorkerShouldFail()
```
lemaitre-aneo, line 581: This should most likely be merged with the previous test.

lemaitre-aneo: The only difference seems to be:
```c#
    Assert.AreEqual(TaskStatus.Submitted,
                    (await testServiceProvider.TaskTable.GetTaskStatus(new[]
                                                                       {
                                                                         tuple.taskSubmitted,
                                                                       },
                                                                       CancellationToken.None)
                                              .ConfigureAwait(false)).Single()
                                                                     .Status);
```

I believe this test should also be valid on all the other cases.

lemaitre-aneo (approved):

PR #172, opened 2022-11-02
Gui endpoint authentication support

dbrasseur-aneo:
- Added authentication support for sessions, tasks and result services used by Gui
- Fixed MongoDB adapter regarding ObjectIds for authentication
- Revamped permissions code to be more generic

### DIFF
--- a/Common/tests/Helpers/GrpcSubmitterServiceHelper.cs
+++ b/Common/tests/Helpers/GrpcSubmitterServiceHelper.cs
@@ -27,7 +27,6 @@ using System.Collections.Generic;
 using System.Net.Http;
 using System.Threading.Tasks;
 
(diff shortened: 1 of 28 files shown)


Common/tests/Helpers/GrpcSubmitterServiceHelper.cs
```diff
@@ -66,6 +70,9 @@ public class GrpcSubmitterServiceHelper : IDisposable
 
     builder.Services.AddSingleton(loggerFactory_)
            .AddSingleton(submitter)
+           .AddSingleton(taskTable)
```
aneojgurhem, line 73: There is no side effects for that ?

aneojgurhem:
The Simple tables are really simple^^

aneojgurhem (approved):

PR #173, opened 2022-11-02
add just script to ease local deployment

jfonseca-aneo:
This script leverages on the functionalities of just. See the installation instructions in [just](https://github.com/casey/just).

Closes https://github.com/aneoconsulting/ArmoniK/issues/683

### DIFF
--- a/tools/justfile
+++ b/tools/justfile
@@ -79,7 +79,8 @@ _usage:
   EOF
 
 # Custom command to build a single image
-build imageTag dockerFile:
+build $imageTag $dockerFile:
+  docker build -t "$imageTag" "$dockerFile" ../
   docker build -t {{ imageTag }} -f {{ dockerFile }} ../
 
 # Build all images necessary for the deployment 
--- a/tools/justfile
+++ b/tools/justfile
@@ -129,6 +129,10 @@ healthChecks:
 remove-dangling:
   docker images --quiet --filter=dangling=true | xargs --no-run-if-empty docker rmi
 
+# Remove deployment images
+remove:
+  docker rmi -f "$ARMONIK_WORKER" "$ARMONIK_METRICS" "$ARMONIK_PARTITIONMETRICS" "$ARMONIK_SUBMITTER" "$ARMONIK_POLLINGAGENT"
+
 # Destroy deployment with docker-compose down
 destroy: (compose "down")
 
--- a/tools/justfile
+++ b/tools/justfile
@@ -1,3 +1,6 @@
+# Enable positional args
+set positional-arguments
+
 # Default values for the deployment
 tag    := "test"
 queue  := "active"
@@ -10,7 +13,7 @@ export COMPOSE_BASE := "../docker-compose/docker-compose.yml"
 export LOGGING_LEVEL := "Information"
 
 # Sets the queue
-export QUEUE := if queue == "rabbit"{
+export QUEUE := if queue == "rabbit" {
   "../docker-compose/docker-compose.queue-rabbitmq.yml"
 }else if queue == "rabbit091"{
   "../docker-compose/docker-compose.queue-rabbitmq.yml"
@@ -22,7 +25,7 @@ export QUEUE := if queue == "rabbit"{
 
 
 # Sets the override to feed docker-compose
-export OVERRIDE := if queue == "rabbit091"{
+export OVERRIDE := if queue == "rabbit091" {
   "../docker-compose/docker-compose.override-rabbitmq091.yml"
 }else{
   "../docker-compose/docker-compose.override.yml"
@@ -100,19 +103,16 @@ set-partitions:
 
 # Custom compose generic rule
 compose *args:
-  docker-compose -f $COMPOSE_BASE -f $OVERRIDE -f $QUEUE {{args}}
+  docker-compose -f $COMPOSE_BASE -f $OVERRIDE -f $QUEUE "$@"
 
 # Call custom docker-compose
-compose-invoke serviceName: (compose "rm -f -s " + serviceName)
+compose-invoke serviceName: (compose "rm" "-f" "-s" serviceName)
 
 # Call custom docker-compose up
-compose-up: (compose "--compatibility up -d --build --force-recreate --remove-orphans")
+compose-up: (compose "--compatibility" "up" "-d" "--build" "--force-recreate" "--remove-orphans")
 
 # Deploy ArmoniK Core
 deploy: build-all
-  #!/usr/bin/env bash
-  set -euo pipefail
-  
   just compose-invoke database
   just compose-invoke queue
   just compose-invoke object
@@ -141,4 +141,4 @@ remove-dangling:
 destroy: (compose "down")
 
 # Custom command to restart the given service
-restart serviceName: (compose "restart " + serviceName)
+restart serviceName: (compose "restart" serviceName)
--- a/tools/justfile
+++ b/tools/justfile
@@ -15,32 +15,38 @@ export LOGGING_LEVEL := "Information"
 # Sets the queue
 export QUEUE := if queue == "rabbit" {
   "../docker-compose/docker-compose.queue-rabbitmq.yml"
-}else if queue == "rabbit091"{
+} else if queue == "rabbit091" {
   "../docker-compose/docker-compose.queue-rabbitmq.yml"
-}else if queue == "artemis"{
+} else if queue == "artemis" {
   "../docker-compose/docker-compose.queue-artemis.yml"
-}else{
+} else {
   "../docker-compose/docker-compose.queue-activemqp.yml"
 }
-
 
 # Sets the override to feed docker-compose
 export OVERRIDE := if queue == "rabbit091" {
   "../docker-compose/docker-compose.override-rabbitmq091.yml"
-}else{
+} else {
   "../docker-compose/docker-compose.override.yml"
 }
 
 # Defines worker and enviroment variables for deployment
 export ARMONIK_WORKER := if worker == "stream" {
-  "dockerhubaneo/armonik_core_stream_test_worker:"+tag
-}else{
-  "dockerhubaneo/armonik_core_htcmock_test_worker:"+tag
+  "dockerhubaneo/armonik_core_stream_test_worker:" + tag
+} else {
+  "dockerhubaneo/armonik_core_htcmock_test_worker:" + tag
 }
-export ARMONIK_METRICS          := "dockerhubaneo/armonik_control_metrics:"+tag
-export ARMONIK_PARTITIONMETRICS := "dockerhubaneo/armonik_control_partitionmetrics:"+tag
-export ARMONIK_SUBMITTER        := "dockerhubaneo/armonik_control:"+tag
-export ARMONIK_POLLINGAGENT     := "dockerhubaneo/armonik_pollingagent:"+tag
+export ARMONIK_METRICS          := "dockerhubaneo/armonik_control_metrics:" + tag
+export ARMONIK_PARTITIONMETRICS := "dockerhubaneo/armonik_control_partitionmetrics:" + tag
+export ARMONIK_SUBMITTER        := "dockerhubaneo/armonik_control:" + tag
+export ARMONIK_POLLINGAGENT     := "dockerhubaneo/armonik_pollingagent:" + tag
+
+# Pick the correct docker file for the worker
+export WORKER_DOCKER_FILE := if worker == "stream" {
+  "../Tests/Stream/Server/Dockerfile"
+} else {
+  "../Tests/HtcMock/Server/src/Dockerfile"
+}
 
 # List recipes and their usage
 @default:
@@ -78,23 +84,11 @@ build imageTag dockerFile:
 
 # Build all images necessary for the deployment 
 build-all:
-  #!/usr/bin/env bash
-  set -euo pipefail
-  if [ "x{{ worker }}" == "xstream" ]
-  then
-    just build $ARMONIK_WORKER  "../Tests/Stream/Server/Dockerfile"
-  elif [ "x{{ worker }}" == "xhtcmock" ]
-  then
-    just build $ARMONIK_WORKER "../Tests/HtcMock/Server/src/Dockerfile"
-  fi
-
-  just build $ARMONIK_METRICS  "../Control/Metrics/src/Dockerfile"
-
-  just build $ARMONIK_PARTITIONMETRICS "../Control/PartitionMetrics/src/Dockerfile"
-
-  just build $ARMONIK_SUBMITTER "../Control/Submitter/src/Dockerfile"
-
-  just build $ARMONIK_POLLINGAGENT "../Compute/PollingAgent/src/Dockerfile"
+  just build "$ARMONIK_WORKER" "$WORKER_DOCKER_FILE"
+  just build "$ARMONIK_METRICS" "../Control/Metrics/src/Dockerfile"
+  just build "$ARMONIK_PARTITIONMETRICS" "../Control/PartitionMetrics/src/Dockerfile"
+  just build "$ARMONIK_SUBMITTER" "../Control/Submitter/src/Dockerfile"
+  just build "$ARMONIK_POLLINGAGENT" "../Compute/PollingAgent/src/Dockerfile"
 
 # Insert partitions in database
 set-partitions:
@@ -103,7 +97,7 @@ set-partitions:
 
 # Custom compose generic rule
 compose *args:
-  docker-compose -f $COMPOSE_BASE -f $OVERRIDE -f $QUEUE "$@"
+  docker-compose -f "$COMPOSE_BASE" -f "$OVERRIDE" -f "$QUEUE" "$@"
 
 # Call custom docker-compose
 compose-invoke serviceName: (compose "rm" "-f" "-s" serviceName)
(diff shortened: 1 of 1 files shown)


lemaitre-aneo (commented):
I think it is better to use variables instead of parameters for tag, queue, and worker. I would also think it is better to use sub-rules instead of making a very large rule.

tools/justfile
```diff
@@ -0,0 +1,156 @@
...
+    QUEUE="../docker-compose/docker-compose.queue-artemis.yml"
+  else
+    echo "{{ queue }} is not a valid queue"
+    exit 1
+  fi
+  echo "export QUEUE=$QUEUE" >> justcompose.env
+  echo "export OVERRIDE=$OVERRIDE" >> justcompose.env
+
+# Defines worker and enviroment variables for deployment
+_set-vars:
```
lemaitre-aneo, line 66: This should be done in `just` variables, and skip the `justcompose.env` file.

jfonseca-aneo: I agree with you, but just doesn't support this: https://github.com/casey/just#setting-variables-in-a-recipe

lemaitre-aneo: You don't need a recipe, you can use a if-then-else outside a rule, at the place you define the variable.

jfonseca-aneo: Ah right!  that will be much nicer! I'm on it

tools/justfile
```diff
@@ -0,0 +1,156 @@
...
+    docker build -t $ARMONIK_WORKER -f ../Tests/HtcMock/Server/src/Dockerfile ../
+  else
+    echo "{{ worker }} is not a valid worker"
+    exit 1
+  fi
+  echo "export ARMONIK_WORKER=$ARMONIK_WORKER" >> justcompose.env
+
+  export ARMONIK_METRICS="dockerhubaneo/armonik_control_metrics:{{ tag }}"
+  #docker rmi -f $ARMONIK_METRICS
+  docker build -t $ARMONIK_METRICS -f ../Control/Metrics/src/Dockerfile ../
```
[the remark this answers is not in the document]
jfonseca-aneo: ok

tools/justfile
```diff
@@ -0,0 +1,144 @@
...
+
+# Sets the override to feed docker-compose
+export OVERRIDE := if  queue == "rabbit091" {
+  "../docker-compose/docker-compose.override-rabbitmq091.yml"
+}else{
+  "../docker-compose/docker-compose.override.yml"
+}
+
+# Defines worker and enviroment variables for deployment
+export ARMONIK_WORKER:= if worker == "stream" {
```
lemaitre-aneo, line 32: Be careful with syntax. It is a bit weird here, and inconsistent

tools/justfile
```diff
@@ -0,0 +1,144 @@
...
+  EOF
+  echo
+  echo "   worker: allowed values below"
+  cat <<-EOF
+        htcmock: for HtcMock V3 (default)
+        stream: for Stream worker
+  EOF
+
+# Custom command to build a single image
+_build-image imageTag dockerFile:
```
lemaitre-aneo, line 73: rename the rule to `build` ?

tools/justfile
```diff
@@ -0,0 +1,144 @@
...
+  docker build -t {{ imageTag }} -f {{ dockerFile }} ../
+
+# Build all images necessary for the deployment 
+_build-armonik-components:
```
lemaitre-aneo, line 77: rename the rule to `build-all`?
I also think you can call subrule at the right of the colon, instead of calling `just` inside.
```suggestion
_build-armonik-components: (build ARMONIK_WORKER "../Tests/Stream/Server/Dockerfile") (build ARMONIK_METRICS "../Control/Metrics/src/Dockerfile")...
```

tools/justfile
```diff
@@ -0,0 +1,144 @@
...
+
+  just _build-image $ARMONIK_SUBMITTER "../Control/Submitter/src/Dockerfile"
+
+  just _build-image $ARMONIK_POLLINGAGENT "../Compute/PollingAgent/src/Dockerfile" 
+
+# Insert partitions in database
+_set-partitions:
+  docker run --net armonik-backend --rm rtsp/mongosh mongosh mongodb://database:27017/database \
+    --eval 'db.PartitionData.insertMany([{ _id: "TestPartition0", ParentPartitionIds: [], PodReserved: 50, PodMax: 100, PreemptionPercentage: 20, Priority: 1, PodConfiguration: null},{ _id: "TestPartition1", ParentPartitionIds: [], PodReserved: 50, PodMax: 100, PreemptionPercentage: 20, Priority: 1, PodConfiguration: null},{ _id: "TestPartition2", ParentPartitionIds: [], PodReserved: 50, PodMax: 100, PreemptionPercentage: 20, Priority: 1, PodConfiguration: null}])'
+
```
lemaitre-aneo, line 100: Add a `compose` generic rule:

```
compose *args:
  docker-compose -f $COMPOSE_BASE -f $OVERRIDE -f $QUEUE {{args}}
```

This generic rule can be called to build more complex rules

tools/justfile
jfonseca-aneo: Which syntax exactly? the `foo := if condition { value_true } {  value_false }`  is the provided in `just` documentation

tools/justfile
jfonseca-aneo: Agree with the renaming, regarding subruling after the colon I find more readable to call `just` it inside.

tools/justfile
lemaitre-aneo: No, I mean the inconsistency with the spaces around the assignment operator. (and around `else`, also)

tools/justfile
lemaitre-aneo: fair enough

tools/justfile
jfonseca-aneo: Ok with the renaming. A question: we let the rule public, also not `_build`?

tools/justfile
jfonseca-aneo: I see, ok i'll take care of it.

tools/justfile
lemaitre-aneo: yes, public is fine. There is no reason to make it private.

tools/justfile
```diff
@@ -70,56 +70,58 @@ _usage:
...
 
 # Insert partitions in database
-_set-partitions:
+set-partitions:
   docker run --net armonik-backend --rm rtsp/mongosh mongosh mongodb://database:27017/database \
     --eval 'db.PartitionData.insertMany([{ _id: "TestPartition0", ParentPartitionIds: [], PodReserved: 50, PodMax: 100, PreemptionPercentage: 20, Priority: 1, PodConfiguration: null},{ _id: "TestPartition1", ParentPartitionIds: [], PodReserved: 50, PodMax: 100, PreemptionPercentage: 20, Priority: 1, PodConfiguration: null},{ _id: "TestPartition2", ParentPartitionIds: [], PodReserved: 50, PodMax: 100, PreemptionPercentage: 20, Priority: 1, PodConfiguration: null}])'
 
+# Custom compose generic rule
+compose *args:
+  docker-compose -f $COMPOSE_BASE -f $OVERRIDE -f $QUEUE {{args}}
```
lemaitre-aneo, line 103: I just checked the doc, and it is not the right way to handle args: https://github.com/casey/just#positional-arguments
```suggestion
  docker-compose -f $COMPOSE_BASE -f $OVERRIDE -f $QUEUE "$@"
```

tools/justfile
```diff
@@ -70,56 +70,58 @@ _usage:
...
+
 # Call custom docker-compose
-_compose-invoke serviceName:
-  docker-compose -f $COMPOSE_BASE -f $OVERRIDE -f $QUEUE rm -f -s {{ serviceName }}
+compose-invoke serviceName: (compose "rm -f -s " + serviceName)
```
lemaitre-aneo, line 106: Each argument should be in its own string
```suggestion
compose-invoke serviceName: (compose "rm" "-f" "-s" serviceName)
```

tools/justfile
```diff
@@ -70,56 +70,58 @@ _usage:
...
 
 # Call custom docker-compose up
-_compose-up:
-  docker-compose -f $COMPOSE_BASE -f $OVERRIDE -f $QUEUE --compatibility up -d --build --force-recreate --remove-orphans
+compose-up: (compose "--compatibility up -d --build --force-recreate --remove-orphans")
```
lemaitre-aneo, line 109: Same here
```suggestion
compose-up: (compose "--compatibility" "up" "-d" "--build" "--force-recreate" "--remove-orphans")
```

tools/justfile
```diff
@@ -70,56 +70,58 @@ _usage:
...
 
 # Deploy ArmoniK Core
-deploy: _build-armonik-components
+deploy: build-all
   #!/usr/bin/env bash
```
lemaitre-aneo, line 113: I am pretty sure you don't need to be in a single bash script

tools/justfile
```diff
@@ -7,37 +7,37 @@ worker := "htcmock"
...
 # loggin level of deployment
-export LOGGING_LEVEL:="Information"
+export LOGGING_LEVEL := "Information"
 
 # Sets the queue
-export QUEUE := if queue == "rabbit" {
+export QUEUE := if queue == "rabbit"{
   "../docker-compose/docker-compose.queue-rabbitmq.yml"
-}else if  queue == "rabbit091" {
+}else if queue == "rabbit091"{
```
lemaitre-aneo, line 15: I would prefer spaces around braces.

lemaitre-aneo (commented):
Close to the end ;)

tools/justfile
```diff
@@ -0,0 +1,144 @@
...
+
+# loggin level of deployment
+export LOGGING_LEVEL := "Information"
+
+# Sets the queue
+export QUEUE := if queue == "rabbit" {
+  "../docker-compose/docker-compose.queue-rabbitmq.yml"
+}else if queue == "rabbit091"{
+  "../docker-compose/docker-compose.queue-rabbitmq.yml"
+}else if queue == "artemis"{
```
lemaitre-aneo, line 20: Could you add whitespaces?
```suggestion
} else if queue == "artemis" {
```

tools/justfile
```diff
@@ -0,0 +1,144 @@
...
+
+# Custom command to build a single image
+build imageTag dockerFile:
+  docker build -t {{ imageTag }} -f {{ dockerFile }} ../
+
+# Build all images necessary for the deployment 
+build-all:
+  #!/usr/bin/env bash
+  set -euo pipefail
+  if [ "x{{ worker }}" == "xstream" ]
```
lemaitre-aneo, line 83: This can also be handled directly with a `just` variable

tools/justfile
```diff
@@ -0,0 +1,144 @@
...
+  just build $ARMONIK_POLLINGAGENT "../Compute/PollingAgent/src/Dockerfile"
+
+# Insert partitions in database
+set-partitions:
+  docker run --net armonik-backend --rm rtsp/mongosh mongosh mongodb://database:27017/database \
+    --eval 'db.PartitionData.insertMany([{ _id: "TestPartition0", ParentPartitionIds: [], PodReserved: 50, PodMax: 100, PreemptionPercentage: 20, Priority: 1, PodConfiguration: null},{ _id: "TestPartition1", ParentPartitionIds: [], PodReserved: 50, PodMax: 100, PreemptionPercentage: 20, Priority: 1, PodConfiguration: null},{ _id: "TestPartition2", ParentPartitionIds: [], PodReserved: 50, PodMax: 100, PreemptionPercentage: 20, Priority: 1, PodConfiguration: null}])'
+
+# Custom compose generic rule
+compose *args:
+  docker-compose -f $COMPOSE_BASE -f $OVERRIDE -f $QUEUE "$@"
```
lemaitre-aneo, line 106: Add quotes around bash variables
```suggestion
  docker-compose -f "$COMPOSE_BASE" -f "$OVERRIDE" -f "$QUEUE" "$@"
```

tools/justfile
```diff
@@ -0,0 +1,144 @@
...
+# Sets the override to feed docker-compose
+export OVERRIDE := if queue == "rabbit091" {
+  "../docker-compose/docker-compose.override-rabbitmq091.yml"
+}else{
+  "../docker-compose/docker-compose.override.yml"
+}
+
+# Defines worker and enviroment variables for deployment
+export ARMONIK_WORKER := if worker == "stream" {
+  "dockerhubaneo/armonik_core_stream_test_worker:"+tag
```
lemaitre-aneo, line 36: I think you can add whitespaces around `+`
```suggestion
  "dockerhubaneo/armonik_core_stream_test_worker:" + tag
```

lemaitre-aneo (approved):
Those last comments are mostly nitpicks. This PR looks good to me otherwise.

tools/justfile
```diff
@@ -0,0 +1,144 @@
...
+
+# Sets the override to feed docker-compose
+export OVERRIDE := if queue == "rabbit091" {
+  "../docker-compose/docker-compose.override-rabbitmq091.yml"
+}else{
+  "../docker-compose/docker-compose.override.yml"
+}
+
+# Defines worker and enviroment variables for deployment
+export ARMONIK_WORKER := if worker == "stream" {
```
aneojgurhem, line 35: What if I have another one that is not there ?

jfonseca-aneo: That case is not handled. A possible solution could be to load a `.env` file specifying the two variables `ARMONIK_WORKER` and `WORKER_DOCKER_FILE`, is such file is not provided we set them in the script.

aneojgurhem: Isn't it possible to do so with command parameters ?

tools/justfile
```diff
@@ -117,9 +111,7 @@ deploy: build-all
   just compose-invoke queue
   just compose-invoke object
   just compose-invoke seq
-
   just compose-up
-
   wait
```
lemaitre-aneo, line 115: What does it is meant to do?

jfonseca-aneo: To make sure mongo is deployed before trying to populate the Db, but it is not necessary, I just removed it

tools/justfile
jfonseca-aneo: I'll re-check the doc and see

tools/justfile
```diff
@@ -0,0 +1,128 @@
...
+        htcmock: for HtcMock V3 (default)
+        stream: for Stream worker
+  EOF
+
+# Custom command to build a single image
+build imageTag dockerFile:
+  docker build -t {{ imageTag }} -f {{ dockerFile }} ../
+
+# Build all images necessary for the deployment 
+build-all: (build "$ARMONIK_WORKER" "$WORKER_DOCKER_FILE") (build "$ARMONIK_METRICS" "../Control/Metrics/src/Dockerfile") (build "$ARMONIK_PARTITIONMETRICS" "../Control/PartitionMetrics/src/Dockerfile") (build "$ARMONIK_SUBMITTER" "../Control/Submitter/src/Dockerfile") (build "$ARMONIK_POLLINGAGENT" "../Compute/PollingAgent/src/Dockerfile")
```
lemaitre-aneo, line 86: I think you can just remove the quotes and the dollar, because those are just variables

tools/justfile
```diff
@@ -0,0 +1,128 @@
...
+  EOF
+  echo
+  echo "   worker: allowed values below"
+  cat <<-EOF
+        htcmock: for HtcMock V3 (default)
+        stream: for Stream worker
+  EOF
+
+# Custom command to build a single image
+build imageTag dockerFile:
```
lemaitre-aneo, line 82: I think you can take the parameters by env (with a dollar), and use env variables in the command:
```suggestion
build $imageTag $dockerFile:
  docker build -t "$imageTag" "$dockerFile" ../
```

tools/justfile
jfonseca-aneo: ok

tools/justfile
lemaitre-aneo: I think that you can change the value of the variables directly in cli if you want.

jfonseca-aneo: just pushed a solution with the env vars

tools/justfile
```diff
@@ -0,0 +1,136 @@
+# Enable positional args
+set positional-arguments
```
lemaitre-aneo, line 2: I am not sure this is required.

tools/justfile
```diff
@@ -0,0 +1,136 @@
...
+  echo
+  echo "The recipe deploy uses three variables"
+  echo
+  echo "usage: just tag=<tag> queue=<queue> worker=<worker> deploy"
+  echo "       if any of the variables is not set, its default value is used"
+  echo
+  echo "   tag: The core tag image to use, defaults to test"
+  echo
+  echo "   queue: allowed values below"
+  cat <<-EOF
```
lemaitre-aneo, line 69: Either all `echo`, or a single `cat <<-EOF``, but the mixing of both is a bit weird

lemaitre-aneo (approved):

tools/justfile
lemaitre-aneo: That nice like this.

tools/justfile
jfonseca-aneo: Yes, its false by default https://github.com/casey/just#table-of-settings

lemaitre-aneo: Oh, I thought `*args` in a rule would be enough.

PR #178, opened 2022-11-09
Some improvements to the pipeline

aneojgurhem:
This PR improve the pipeline

- [x] It reduces the number of jobs by building the solution containing the projects in one go instead of using a job by project. Now, there is one job that takes less than a minute and build all the projects instead of 20+ jobs that take around 30s.
- [x] It mutualises more executions of the HTC Mock in the same test job and run the large jobs only in Information log level.
- [x] Some of the dependencies were reworked to build the docker images earlier.

fix https://github.com/aneoconsulting/ArmoniK/issues/696

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -503,13 +503,13 @@ jobs:
           - activemqp
           - rabbitmq
           - rabbitmq091
-        log-level:
-          - Information
-          - Verbose
-        ntasks:
-          - 100
-          - 1000
-          - 1500
+        include:
+          - log-level: Verbose
+            ntasks: 100
+          - log-level: Information
+            ntasks: 100
+          - log-level: Information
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -537,10 +506,7 @@ jobs:
         log-level:
           - Information
           - Verbose
-        ntasks:
-          - 100
-          - 1000
-          - 1500
```
jfonseca-aneo, line 543: I'd keep the `ntasks` matrix, even if it has only one value it will make easier to customize the test in the future if needed

aneojgurhem: I removed it to reduce the number of jobs. They are now executed in the same job. Although the 1000 is only executed in Information log level

jfonseca-aneo (approved):

PR #179, opened 2022-11-10
Local storage for payloads

lemaitre-aneo:
This PR adds a new adapter to store payloads in a local filesystem. This is intended to replace Redis in case of local deployment.

### DIFF
--- a/Adaptors/LocalStorage/src/ObjectStorage.cs
+++ b/Adaptors/LocalStorage/src/ObjectStorage.cs
@@ -31,6 +31,7 @@ using System.Threading;
 using System.Threading.Tasks;
 
 using ArmoniK.Api.Common.Utils;
+using ArmoniK.Core.Common.Exceptions;
 using ArmoniK.Core.Common.Storage;
 
 using Microsoft.Extensions.Logging;
@@ -42,7 +43,6 @@ public class ObjectStorage : IObjectStorage
   private readonly int                    chunkSize_;
   private readonly ILogger<ObjectStorage> logger_;
   private readonly string                 path_;
-  private readonly int                    pid_;
 
   /// <summary>
   ///   <see cref="IObjectStorage" /> implementation for LocalStorage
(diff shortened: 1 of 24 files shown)


lemaitre-aneo:
With (almost) empty payloads and results, `LocalStorage` is consistently a tiny bit faster than `Redis` (~4%).

Adaptors/LocalStorage/src/ObjectStorage.cs
```diff
@@ -0,0 +1,168 @@
...
+    {
+      // We need to await the read task before starting to read another chunk
+      var read       = await readTask.ConfigureAwait(false);
+      var readBuffer = buffer;
+
+      if (read == 0)
+      {
+        yield break;
+      }
+
```
aneojgurhem, line 135: you should be able to put that in the while

lemaitre-aneo: I don't a see way to do it without duping most of the loop body.

aneojgurhem:
```suggestion
    int read;
    while ((read = await readTask.ConfigureAwait(false)) == 0)
    {
      var readBuffer = buffer;

```

aneojgurhem: Moreover, you can do things after the `yield return` but before the end of the while block.

Control/Submitter/src/ArmoniK.Core.Control.Submitter.csproj
```diff
@@ -29,23 +29,24 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Grpc.AspNetCore.Web" Version="2.48.0" />
-    <PackageReference Include="Microsoft.VisualStudio.Azure.Containers.Tools.Targets" Version="1.17.0" />
-    <PackageReference Include="OpenTelemetry.Exporter.Zipkin" Version="1.3.1" />
-    <PackageReference Include="OpenTelemetry.Extensions.Hosting" Version="1.0.0-rc9.6" />
-    <PackageReference Include="OpenTelemetry.Instrumentation.AspNetCore" Version="1.0.0-rc9.6" />
```
aneojgurhem, line 36: This means that the pipeline does not check if the csproj files are correctly formatted

aneojgurhem (approved):

PR #181, opened 2022-11-10
Enable  healthcheck  for polling agent in docker compose

jfonseca-aneo:
Fix https://github.com/aneoconsulting/ArmoniK/issues/623


docker-compose/docker-compose.yml
```diff
@@ -146,6 +146,13 @@ services:
       driver: fluentd
       options:
         fluentd-address: 127.0.0.1:24224
+    healthcheck:
+      test: ["CMD", "fcurl.sh", "localhost", "1080", "/readiness"]
```
lemaitre-aneo, line 150: You can directly "write" the script into the health check command. This would avoid the need to change the image.

```suggestion
      test:
      - "CMD"
      - "bash"
      - "-c"
      - |
        {
          echo -en "GET /liveness HTTP/1.1\r\nHost: localhost:80\r\nConnection: close\r\n\r\n">&3 &
          read content <&3
          wait
        } 3<>/dev/tcp/localhost/80
        case "$content" in
          "HTTP/1.1 200"*)
            echo "OK"
            exit 0
            ;;
        esac
        echo "KO" >&2
        exit 1
```

jfonseca-aneo: Ah yes!  `read` never defined `content` for me so I tried another trick

aneojgurhem:
@lemaitre-aneo @jfonseca-aneo with terraform coming, do we include it or close it ?

jfonseca-aneo:
You might close it, we can include the health check in the terraform itself.

PR #182, opened 2022-11-10
Use just script to deploy in the pipeline

jfonseca-aneo:
Fix https://github.com/aneoconsulting/ArmoniK/issues/686


docker-compose/docker-compose.yml
```diff
@@ -52,6 +52,19 @@ services:
...
       options:
         fluentd-address: 127.0.0.1:24224
+    healthcheck:
+      test:
+        - "CMD"
+        - "bash"
+        - "-c"
+        - |
+          exec 3<>"/dev/tcp/localhost/1081"
+          echo -e "GET /liveness HTTP/1.1\r\nHost: localhost:1081\r\nConnection: close\r\n\r\n">&3 &
```
lemaitre-aneo, line 62: `echo -en` instead: disable extra line return

docker-compose/docker-compose.yml
```diff
@@ -52,6 +52,19 @@ services:
...
+          cat <&3 | grep Healthy || exit 1 && echo "$$?"
```
lemaitre-aneo, line 63:
```suggestion
          if grep Healthy <&3 &>/dev/null ; then
            echo "$?"
          else
            exit 1
          fi
```

BTW, why do you have `"$$?"

lemaitre-aneo: Also, why do you do this echo?

jfonseca-aneo: I was just a really dirty way to get to emulate the `if` condition, (I had no idea I can "hardcode" bash code in the .yml in the way you suggested ^^). It's no longer necessary

lemaitre-aneo: Ok so you can just keep the grep only:
```suggestion
          grep Healthy <&3 &>/dev/null
```

And be sure that it is the last command

jfonseca-aneo: Docker compose expects either zero or one as output of your test if you feed it with other values, (in particular with a two) it might do unexpected stuff crazy

lemaitre-aneo: Tru, I forgot about that. So jsut add `|| exit 1` and nothing else

aneojgurhem (approved):

lemaitre-aneo (approved):

PR #184, opened 2022-11-17
Implement events API

aneojgurhem:
This PR adds the implementation for the service called Events. This service can be used to receive events related to the update of tasks and results whithin a session. The endpoint can be called to listen to the modifications of multiple sessions if needed. These events can be used to form the graph of data dependencies between the tasks of a session.

As for now, all the events of a session will be sent whithout filtering. It is possible that the API will evolve to a more refined way to filter the events to be received.

- [x] Create interface and write code to listen to mongodb change stream (replication logs)
- [x] Implement gRPC service based on the code to listen to change stream
- [x] Create a mongodb cluster with one replica for testing in docker compose
- [x] Tests
- [x] write doc
- [x] This feature does not break ArmoniK when there is no mongodb cluster. If there is no mongodb cluster, the endpoint does not work properly, but the other endpoints are working as before.

Note: https://hackmd.io/HnFWBFuSTbSdwlyqo6fGgQ?view

fix https://github.com/aneoconsulting/ArmoniK/issues/724


Adaptors/MongoDB/src/ChangeStreamUpdate.cs
```diff
@@ -0,0 +1,94 @@
...
+                                   "tmpfields.k", new BsonDocument
+                                                  {
+                                                    {
+                                                      "$in", new BsonArray(fields)
+                                                    },
+                                                  }
+                                 },
+                               }
+                   },
+                 };
```
dbrasseur-aneo, line 75: Can surely be transformed into a Linq-style expression

aneojgurhem: I wish it was

Common/tests/ResultWatcherTestBase.cs
```diff
@@ -0,0 +1,326 @@
...
+                      (await ResultWatcher.Check(HealthCheckTag.Readiness)
+                                          .ConfigureAwait(false)).Status);
+      Assert.AreEqual(HealthStatus.Healthy,
+                      (await ResultWatcher.Check(HealthCheckTag.Startup)
+                                          .ConfigureAwait(false)).Status);
+    }
+  }
+
+
+  private static async Task ProduceEvents(IResultTable      resultTable,
```
jfonseca-aneo, line 165: It will be really useful to have some text specifying which events this function generate, and then at each test, how the events are updated and the expected behavior of the test (even if this is implicitly stated in the test name)

Common/tests/TaskWatcherTestBase.cs
```diff
@@ -0,0 +1,396 @@
...
+      Assert.AreEqual(HealthStatus.Healthy,
+                      (await TaskWatcher.Check(HealthCheckTag.Readiness)
+                                        .ConfigureAwait(false)).Status);
+      Assert.AreEqual(HealthStatus.Healthy,
+                      (await TaskWatcher.Check(HealthCheckTag.Startup)
+                                        .ConfigureAwait(false)).Status);
+    }
+  }
+
+
```
jfonseca-aneo, line 266: Idem

Adaptors/MongoDB/src/WatchEnumerator.cs
```diff
@@ -0,0 +1,112 @@
...
+using System.Collections.Generic;
+using System.Threading;
+
+using ArmoniK.Core.Common.Storage;
+
+using MongoDB.Driver;
+
+namespace ArmoniK.Core.Adapters.MongoDB;
+
+public sealed class WatchEnumerator<TOutput, TInput> : IWatchEnumerator<TOutput>
```
jfonseca-aneo, line 35: Documentation: What it the relation between a `cursor` and a `currentEnumerable`

aneojgurhem: a cursor is a IEnumerable<IEnumerable<T>> so you need a bit of code to make it a proper Enumerator

Common/src/gRPC/Services/WatchToGrpc.cs
```diff
@@ -0,0 +1,299 @@
...
+                                                                          page,
+                                                                          PageSize,
+                                                                          cancellationToken)
+                                                          .ConfigureAwait(false)).totalCount > read)
+                            {
+                              using var enumerator = res.tasks.GetEnumerator();
+
+                              while (enumerator.MoveNext())
+                              {
+                                var cur = enumerator.Current;
```
lemaitre-aneo, line 82: I don't think you need to explicitly use the enumerator here:
```suggestion
                              foreach (var cur in res.tasks) {
```

Common/src/gRPC/Services/WatchToGrpc.cs
```diff
@@ -0,0 +1,299 @@
...
+    Task.Factory.StartNew(async () =>
+                          {
+                            var newTasks = await taskWatcher_.GetNewTasks(sessionId,
+                                                                          cancellationToken)
+                                                             .ConfigureAwait(false);
+
+                            while (await newTasks.MoveNextAsync()
+                                                 .ConfigureAwait(false))
+                            {
+                              var cur = newTasks.Current;
```
lemaitre-aneo, line 165: It will be the same here once you have a WatchEnumerable

Adaptors/MongoDB/src/WatchEnumerator.cs
```diff
@@ -0,0 +1,117 @@
...
+  /// <inheritdoc />
+  public TOutput Current
+    => currentEnumerator_ is null
+         ? throw new InvalidOperationException()
+         : converter_(currentEnumerator_.Current);
+
+  /// <inheritdoc />
+  public ValueTask DisposeAsync()
+  {
+    currentEnumerator_?.Dispose();
```
lemaitre-aneo, line 114: The cursor is also disposable, should most likely be disposed here.

Common/src/gRPC/Services/GrpcEventsService.cs
```diff
@@ -0,0 +1,84 @@
...
+                                   context.CancellationToken)
+                        .GetAsyncEnumerator();
+
+    while (await enumerator.MoveNextAsync(context.CancellationToken)
+                           .ConfigureAwait(false))
+    {
+      await responseStream.WriteAsync(enumerator.Current,
+                                      context.CancellationToken)
+                          .ConfigureAwait(false);
+    }
```
lemaitre-aneo, line 82: foreach on the enumerable + try catch on the cancellation to close stream

lemaitre-aneo:
You most likely need also a test that calls the watch multiple times in parallel.

lemaitre-aneo (approved):

PR #192, opened 2022-11-22
Revert "Fix versionning in pipeline"

dbrasseur-aneo:
Fix is no longer needed
This reverts commit ce19b70eb42b11c7a5221ed77f73ffbf89ff5b65.

Should we pick a specific version and stick with it ?

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -72,9 +72,9 @@ jobs:
       run: |
         if [[ "${{ github.ref_name }}" == "releases/"* ]]
         then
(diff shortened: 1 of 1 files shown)


lemaitre-aneo:
We don't need to revert on the main branch. But we can revert on the next branch. All in all, I prefer to fix the version, but this might lead to use very old versions in the future.

lemaitre-aneo (approved):

lemaitre-aneo:
@aneojgurhem Do we proceed with this PR?

aneojgurhem:
Now that the issue seems to be fixed, we can use the latest version.

PR #198, opened 2022-12-05
Fix mongo db tests

aneojgurhem:
For an unknown reason, all mongodb tests on Linux were failing during mongo2go initialization.
This PR uses EphemeralMongo instead of Mongo2Go for the tests. It solved the issue.

see https://github.com/Mongo2Go/Mongo2Go/issues/141

fix https://github.com/aneoconsulting/ArmoniK/issues/745

### DIFF
--- a/Common/tests/ArmoniK.Core.Common.Tests.csproj
+++ b/Common/tests/ArmoniK.Core.Common.Tests.csproj
@@ -13,9 +13,9 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="EphemeralMongo5" Version="0.1.3" />
     <PackageReference Include="Microsoft.AspNetCore.TestHost" Version="6.0.9" />
     <PackageReference Include="Microsoft.Azure.Amqp" Version="2.5.12" />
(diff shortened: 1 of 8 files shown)


Common/tests/ArmoniK.Core.Common.Tests.csproj
```diff
@@ -13,9 +13,9 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="EphemeralMongo5" Version="0.1.3" />
```
jfonseca-aneo, line 16: Is there a reason not to pick EphemeralMongo6?

lemaitre-aneo: I believe the infra still uses mongo 5, so better keep it in sync ?

lemaitre-aneo (approved):

Common/tests/ArmoniK.Core.Common.Tests.csproj
aneojgurhem: Yes, the infra uses mongo 5

PR #199, opened 2022-12-06
Put task in error when worker is unavailable during task execution

aneojgurhem:
fix https://github.com/aneoconsulting/ArmoniK/issues/743

### DIFF
--- a/Common/src/Pollster/TaskHandler.cs
+++ b/Common/src/Pollster/TaskHandler.cs
@@ -581,9 +581,9 @@ public class TaskHandler : IAsyncDisposable
     }
     catch (Exception e)
     {
-      await HandleErrorAsync(e,
-                             taskData_,
-                             cancellationTokenSource_.Token)
+      await HandleErrorResubmitAsync(e,
+                                     taskData_,
+                                     cancellationTokenSource_.Token)
         .ConfigureAwait(false);
     }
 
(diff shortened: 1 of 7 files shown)


Common/src/Pollster/TaskHandler.cs
```diff
@@ -589,33 +590,19 @@ await agent_.FinalizeTaskCreation(CancellationToken.None)
     logger_.LogDebug("End Task Processing");
   }
 
-  private async Task HandleErrorAsync(Exception         e,
-                                      TaskData          taskData,
-                                      CancellationToken cancellationToken)
+  private async Task HandleErrorRequeueAsync(Exception         e,
```
lemaitre-aneo, line 593: I think it is a good idea to have 2 separate functions for those tasks, but the implementation is very much similar, so maybe they could call a single function, no?

aneojgurhem: The "unified" function would be my first commit. It increase the complexity of the code too much.

lemaitre-aneo: Just to be clear, my proposal was to have two functions when calling, but those two functions call a single one. From what I can see, the only real difference between the two is the status of the message, and if the task is actually resubmitted or not. Those can be taken in parameter of the helper function.

All in all, I don't mind the current implementation.

Common/src/Pollster/TaskHandler.cs
```diff
@@ -594,10 +594,36 @@ await agent_.FinalizeTaskCreation(CancellationToken.None)
...
+                                   true,
+                                   false,
+                                   cancellationToken)
+      .ConfigureAwait(false);
+  }
+
+  private async Task HandleErrorInternalAsync(Exception         e,
+                                              TaskData          taskData,
+                                              bool              resubmit,
+                                              bool              requeueIfUnavailable,
```
lemaitre-aneo, line 620: `requeueIfUnavalible` does not seem to be used. However, the refactor changed the status of the queue message.

aneojgurhem: It is used.
Good point ! I'll fix it.

Common/src/Pollster/TaskHandler.cs
```diff
@@ -580,42 +581,50 @@ await agent_.FinalizeTaskCreation(CancellationToken.None)
...
         .ConfigureAwait(false);
     }
 
     logger_.LogDebug("End Task Processing");
   }
 
-  private async Task HandleErrorAsync(Exception         e,
-                                      TaskData          taskData,
-                                      CancellationToken cancellationToken)
+  private async Task HandleErrorRequeueAsync(Exception         e,
```
lemaitre-aneo, line 593: I wonder if it would be better to make this method just return the task from internal, instead of making this function asynchronous and awaiting it.
This does not block the PR anyway.

lemaitre-aneo (approved):

Common/src/Pollster/TaskHandler.cs
aneojgurhem: I don't want to try the propogation of exceptions in async tasks.

PR #200, opened 2022-12-07
Use terraform to deploy ArmoniK.Core infrastructure

jfonseca-aneo:
A first set of `.tf` files to deploy Core. Tasks still to complete:

- [x] Fix naming convention
- [x] Remove explicit dependencies
- [x] Add support for locally generated images 
- [x] A generic configuration for the ports/sockets to be used.
- [x] A generic configuration for the polling agent
- [x] A generic configuration to choose the worker.
- [x] Enable option for RabbitMQ.
- [x] Enable option for local storage.
- [x] Generate environment variables for the compute/control plane via the submodules responsible of their dependencies
- [x] Configure multiple partitions 

Fix https://github.com/aneoconsulting/ArmoniK/issues/755


terraform/backend.tf
```diff
@@ -0,0 +1,80 @@
...
+    name = docker_network.armonik-backend.name
+  }
+
+  ports {
+    internal = 27017
+    external = 27017
+  }
+}
+
+resource "null_resource" "partitions-in-db" {
```
lemaitre-aneo, line 15: you should use `_` instead of `-` in names

terraform/backend.tf
```diff
@@ -0,0 +1,80 @@
+resource "docker_container" "database" {
+  name  = "database"
+  image = var.database-image
```
lemaitre-aneo, line 3: You should use a `docker_image` resource to pass the hash to the docker container

terraform/backend.tf
```diff
@@ -0,0 +1,80 @@
...
+    internal = 5672
+    external = 5672
+  }
+
+  ports {
+    internal = 8161
+    external = 8161
+  }
+
+  healthcheck {
```
lemaitre-aneo, line 61: Healthcheck is a property of a container? nice!

terraform/main.tf
```diff
@@ -0,0 +1,191 @@
...
+    external = 5001
+  }
+
+  ports {
+    internal = 1081
+    external = 5011
+  }
+
+  depends_on = [
+    docker_container.database,
```
lemaitre-aneo, line 66: prefer referencing those resources in the inputs of the current resource instead of relying on explicit dependencies.

terraform/main.tf
```diff
@@ -0,0 +1,191 @@
...
+
+  log_driver = "fluentd"
+
+  log_opts = {
+    fluentd-address = "127.0.0.1:24224"
+  }
+
+  ports {
+    internal = 1080
+    external = tonumber("108${each.value}")
```
lemaitre-aneo, line 176:
```suggestion
    external = 1080 + each.value
```

terraform/network.tf
```diff
@@ -0,0 +1,11 @@
+resource "docker_network" "armonik-net" {
```
lemaitre-aneo, line 1: Why 3 networks?

terraform/providers.tf
```diff
@@ -0,0 +1,12 @@
...
+  required_providers {
+    docker = {
+      source  = "kreuzwerker/docker"
+      version = "~> 2.23.0"
+    }
+  }
+}
+
+provider "docker" {
+  host = "unix:///var/run/docker.sock"
```
lemaitre-aneo, line 11: remove this line. It is the default configuration on Linux

terraform/network.tf
aneojgurhem: That's what I did in the docker-compose files

jfonseca-aneo: I'm using only one in the last commited version

terraform/modules/localImage/main.tf
```diff
@@ -0,0 +1,12 @@
+resource "null_resource" "build_local_image" {
```
lemaitre-aneo, line 1: There is an option in the docker provider to build the image. You should use that instead.

terraform/volumes.tf
```diff
@@ -1,8 +1,8 @@
 locals {
-  vol-replicas = toset([for s in range(var.num-replicas) : tostring(s)])
+  vol_replicas = toset([for s in range(var.num_replicas) : tostring(s)])
```
lemaitre-aneo, line 2: You don't need tostring here.

jfonseca-aneo: You actually do

terraform/modules/pollingAgent/main.tf
```diff
@@ -0,0 +1,74 @@
+locals {
+  test-cmd = <<-EOF
+    CMD bash -c exec 3<>"/dev/tcp/localhost/1080"
+    echo -en "GET /liveness HTTP/1.1\r\nHost: localhost:1080\r\nConnection: close\r\n\r\n">&3 &
+    grep Healthy <&3 &>/dev/null || exit 1
+    EOF
```
lemaitre-aneo, line 6: Better to just put the script here, and add the boilerplate in the healthcheck block (cf: next comment)

```suggestion
  test_cmd = <<-EOF
    exec 3<>"/dev/tcp/localhost/1080"
    echo -en "GET /liveness HTTP/1.1\r\nHost: localhost:1080\r\nConnection: close\r\n\r\n">&3 &
    grep Healthy <&3 &>/dev/null || exit 1
    EOF
```

terraform/modules/pollingAgent/main.tf
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  mounts {
+    type   = "volume"
+    target = "/cache"
+    source = var.socket_vol
+  }
+
+  healthcheck {
+    test         = split(" ", local.test-cmd)
```
lemaitre-aneo, line 68:
```suggestion
    test         = ["CMD", "bash", "-c", local.test_cmd]
```

terraform/modules/queue/activemq/inputs.tf
```diff
@@ -0,0 +1,12 @@
+variable "image" {
+  type = string
+}
+
+variable "network" {
+  type = string
+}
+
+variable "exposed_ports" {
+  type    = list(number)
```
lemaitre-aneo, line 10: if you need a list with exactly 2 elements, prefer an object.

terraform/main.tf
```diff
@@ -2,69 +2,96 @@ locals {
...
+
+module "queue" {
+  source  = "./modules/queue/activemq"
+  image   = var.queue_image
+  network = docker_network.armonik.name
+}
 
-  implicit_dependencies = [
-  ]
+module "submitter" {
```
lemaitre-aneo, line 41: I would pass the environment variables "generated" by the db and object storage as arguments. This would allow an implicit dependency, and make more modular.

terraform/modules/localImage/main.tf
jfonseca-aneo: I tried this option but it fails because it does not support having the `Dockerfile` located in a different folder than the context

terraform/main.tf
jfonseca-aneo: Agree, I'll work on that

terraform/modules/compute_plane/main.tf
```diff
@@ -0,0 +1,135 @@
...
+    "Amqp__User=admin",
+    "Amqp__Password=admin",
+    "Amqp__Host=queue",
+    "Amqp__Port=5672",
+    "Amqp__Scheme=AMQP",
+    "Amqp__MaxPriority=10",
+    "Amqp__MaxRetries=10",
+    "Amqp__LinkCredit=2",
+    "Amqp__PartitionId=TestPartition${var.replica_counter}"
+  ]
```
lemaitre-aneo, line 109: I think this should be an input of the module, and generated by each dependant modules and merged together. For instance, the module amqp would generate all the entries related to the queue (including the Components__QueueStorage), the redis module would generate the entries related to the object storage...

terraform/locals.tf
```diff
@@ -1,4 +1,7 @@
 locals {
-  submitter     = merge(var.submitter, { queue_storage = var.queue_storage }, { tag = var.core_tag })
-  compute_plane = merge(var.compute_plane, { queue_storage = var.queue_storage }, { tag = var.core_tag })
+  replicas        = toset([for s in range(var.num_replicas) : tostring(s)])
+  queue_env_vars  = var.queue_storage.broker.name == "activemq" ? module.queue_activemq[0].queue_env_vars : module.queue_rabbitmq[0].queue_env_vars
```
lemaitre-aneo, line 3: I would slightly prefer to have this:
```suggestion
  queue = one(concat(module.queue_activemq, module.queue_rabbitmq))
  queue_env_vars = queue.queue_env_vars
```

This would avoid the need for the condition, and would scale better if we have more queues at our disposal.

terraform/main.tf
```diff
@@ -21,39 +17,53 @@ module "zipkin" {
...
   docker_image      = local.submitter.image
   log_level         = local.submitter.log_level
   dev_env           = local.submitter.aspnet_core_env
-  object_storage    = local.submitter.object_storage
   network           = docker_network.armonik.name
   database_env_vars = module.database.database_env_vars
-  queue_env_vars    = module.queue.queue_env_vars
-  object_env_vars   = module.object.object_env_vars
+  queue_env_vars    = local.queue_env_vars
+  object_env_vars   = local.object_env_vars
```
lemaitre-aneo, line 65: I would recommend to do the merge of the environment variables in the root module instead of the main one.
```terraform
environment = merge(local.queue_env_vars, local.object_env_vars, local.database_env_vars, local.logging_env_vars)
```

terraform/modules/monitoring/metrics/inputs.tf
```diff
@@ -15,24 +15,21 @@ variable "network" {
...
+  type = string
 }
 
-variable "db_driver" {
-  type = object({
-    name = string,
-    port = number,
-  })
+variable "database_env_vars" {
+  type = map(any)
```
lemaitre-aneo, line 27:
```suggestion
  type = map(string)
```

terraform/modules/monitoring/metrics/locals.tf
```diff
@@ -0,0 +1,7 @@
+locals {
+  env = [
+    "Serilog__MinimumLevel=${var.log_level}",
+    "ASPNETCORE_ENVIRONMENT=${var.dev_env}"
+  ]
```
lemaitre-aneo, line 5: Should be given by the root module, in a `environment` variable. This would also reduce the number of variables to pass to each modules.

terraform/modules/monitoring/metrics/locals.tf
```diff
@@ -0,0 +1,7 @@
...
+  db_env = [for t in keys(var.database_env_vars) : format("%s=%s", t, lookup(var.database_env_vars, t))]
```
lemaitre-aneo, line 6: This can be simply rewritten like this:
```suggestion
  db_env = [for k, v in var.database_env_vars) : "${k}=${v}"]
```

lemaitre-aneo (dismissed):
This is in very good shape. The remaining comments are nitpicks.

lemaitre-aneo (approved):

PR #203, opened 2022-12-08
Jg/streamtests

aneojgurhem:
- Improves the error message for tasks in error when the result is not available
- Stat to use channel in HtcMock instead of SubmitterClient in order to be able to use other clients.

fix https://github.com/aneoconsulting/ArmoniK/issues/756

### DIFF
--- a/Tests/Stream/Client/GrpcChannelExt.cs
+++ b/Tests/Stream/Client/GrpcChannelExt.cs
@@ -1,0 +1,94 @@
+// This file is part of the ArmoniK project
+// 
+// Copyright (C) ANEO, 2021-2022. All rights reserved.
+//   W. Kirschenmann   <wkirschenmann@aneo.fr>
+//   J. Gurhem         <jgurhem@aneo.fr>
+//   D. Dubuc          <ddubuc@aneo.fr>
+//   L. Ziane Khodja   <lzianekhodja@aneo.fr>
+//   F. Lemaitre       <flemaitre@aneo.fr>
+//   S. Djebbar        <sdjebbar@aneo.fr>
+//   J. Fonseca        <jfonseca@aneo.fr>
+//   D. Brasseur       <dbrasseur@aneo.fr>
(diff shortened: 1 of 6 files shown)


Tests/Stream/Client/GrpcChannelExt.cs
```diff
@@ -0,0 +1,94 @@
...
+
+      var taskClient = new Tasks.TasksClient(channel);
+      var taskData = await taskClient.GetTaskAsync(new GetTaskRequest
+                                                   {
+                                                     TaskId = res.ResultTask.Single()
+                                                                 .TaskId,
+                                                   })
+                                     .ConfigureAwait(false);
+
+      if (taskData.Task.Output.Success)
```
jfonseca-aneo, line 85: I don't understand this line, why to throw on Success?

Tests/Stream/Client/StreamWrapperTests.cs
```diff
@@ -69,13 +71,22 @@ public void SetUp()
...
+    channel_ = GrpcChannelFactory.CreateChannel(options);
+    client_  = new Submitter.SubmitterClient(channel_);
     Console.WriteLine("Client created");
   }
 
+  [TearDown]
+  public void TearDown()
+  {
+    partition_ = null;
+    client_    = null;
```
jfonseca-aneo, line 83: Do we know if `client_` is properly disposed ?

Tests/Stream/Client/GrpcChannelExt.cs
aneojgurhem: We throw the original error (the one coming from the getresultasync)

Tests/Stream/Client/StreamWrapperTests.cs
aneojgurhem: There are no dispose function in it

jfonseca-aneo (approved):

PR #204, opened 2022-12-09
Improve task acquisition when messages are duplicated

aneojgurhem:
This PR improves the acquisition of the task when the message is duplicated which happen a lot with the Artemis message queue.

Several polling agents were able to acquire the same thus leading to lots of issues when 2 tasks are being processed at the same time. Previously the endpoint `/taskprocessing` of the polling agent was returning the task id only after the acquisition of the task. In the time between the reception of the message and the acquisition of the task on the first polling agent, a second one was able to try to acquire the task, receive an empty response from the first polling agent, consider that the other polling agent was not running the task and resubmit the task. Now the endpoint returns the task id in the message from the queue before the acquisition, allowing the other polling agent to know that the task is already being processed and to deduplicate the message.

fix https://github.com/aneoconsulting/ArmoniK/issues/757

### DIFF
--- a/Common/src/Pollster/TaskHandler.cs
+++ b/Common/src/Pollster/TaskHandler.cs
@@ -357,6 +357,8 @@ public class TaskHandler : IAsyncDisposable
           taskData_ = await taskTable_.ReadTaskAsync(messageHandler_.TaskId,
                                                      CancellationToken.None)
                                       .ConfigureAwait(false);
+          logger_.LogInformation("Task is not running on the other polling agent, status : {status}",
+                                 taskData_.Status);
           if (taskData_.Status is TaskStatus.Processing or TaskStatus.Dispatched or TaskStatus.Processed)
           {
             logger_.LogDebug("Resubmitting task {task} on another pod",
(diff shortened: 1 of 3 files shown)


Common/src/Pollster/TaskHandler.cs
```diff
@@ -357,6 +357,8 @@ public async Task<bool> AcquireTask()
           taskData_ = await taskTable_.ReadTaskAsync(messageHandler_.TaskId,
                                                      CancellationToken.None)
                                       .ConfigureAwait(false);
+          logger_.LogInformation("Task is not running on the other polling agent, status : {status}",
```
ngruelaneo, line 360: Does this log belong to the info status? I am not seeing the value for the people which will operate.

lemaitre-aneo (approved):

Common/src/Pollster/TaskHandler.cs
jfonseca-aneo: I agree.

jfonseca-aneo (approved):

Common/src/Pollster/TaskHandler.cs
aneojgurhem: And this log should not happen very often.

PR #206, opened 2022-12-14
update StreamWrapper docker file to avoid to copy sources

alouvion-aneo:


Tests/Stream/Client/Dockerfile
```diff
@@ -9,8 +12,13 @@ RUN dotnet restore "Tests/Stream/Client/ArmoniK.Extensions.Common.StreamWrapper.
...
 RUN dotnet build "ArmoniK.Extensions.Common.StreamWrapper.Tests.Client.csproj" -c Release -o /app/build
+
+
+FROM base as final
+WORKDIR /app
+COPY --from=build /app/build .
 RUN groupadd --gid 5000 armonikuser && useradd --home-dir /home/armonikuser --create-home --uid 5000 --gid 5000 --shell /bin/sh --skel /dev/null armonikuser
-RUN chown -R armonikuser: /src
+RUN chown -R armonikuser: /app
 RUN rm -rf /tmp/NuGetScratch
```
aneojgurhem, line 22: This would be nice to put those commands in the same run to reduce the size of the image.

aneojgurhem (approved):

PR #211, opened 2023-01-12
Al/815 add s3 adapter to store the payload

alouvion-aneo:
fix https://github.com/aneoconsulting/ArmoniK/issues/815
Epic : https://github.com/aneoconsulting/ArmoniK/issues/810
new S3 adapter is added. It can store payloads.
Docker factory updated.
Terraform scripts for newcoming factory in other PR : https://github.com/aneoconsulting/ArmoniK/issues/840


Adaptors/S3/src/Options/S3Storage.cs
```diff
@@ -0,0 +1,35 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+// 
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+namespace ArmoniK.Core.Adapters.S3.Options;
+
+public class S3
```
aneojgurhem, line 27: file name does not match class name

.github/workflows/build.yml
```diff
@@ -100,7 +100,7 @@ jobs:
 
     - name: Install ReSharper
       run: |
-        dotnet tool install -g JetBrains.ReSharper.GlobalTools --version 2022.2.2
+        dotnet tool install -g JetBrains.ReSharper.GlobalTools --version 2022.3.1
```
lemaitre-aneo, line 103: This should not change. If you want to update resharper in the pipeline, do it in a separate PR, but be aware that it can actually change the way code is formatted. That is why it has been hardcoded.

.github/workflows/build.yml
```diff
@@ -148,6 +149,11 @@ jobs:
       with:
         dotnet-version: 6.x
 
+    - name: Minio Server UP
+      uses: isbang/compose-action@v1.4.0
+      with:
+        compose-file: "./docker-compose/S3/docker-compose-minio.yml"
```
lemaitre-aneo, line 155: Could you add Minio to the terraform deployment? We plan to remove docker compose from core in the (not so distant) future.

Adaptors/S3/tests/Usings.cs
```diff
@@ -0,0 +1,25 @@
...
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY, without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+global using NUnit.Framework;
```
lemaitre-aneo, line 25: I am not fond of `global using` and its non local effects. I believe it is the only place were it is used.
Here it is especially not needed as you have a single `.cs` file.

Adaptors/S3/src/ArmoniK.Core.Adapters.S3.csproj
```diff
@@ -0,0 +1,18 @@
...
+
+  <PropertyGroup>
+    <TargetFramework>net6.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="AWSSDK.S3" Version="3.7.101.46" />
+    <PackageReference Include="CommunityToolkit.HighPerformance" Version="8.0.0" />
```
lemaitre-aneo, line 11: I believe you do not use it.

.github/workflows/build.yml
alouvion-aneo: I agree, it was just a test to see if newer version was improving perfs, but not really

Adaptors/S3/tests/Usings.cs
alouvion-aneo: Same for me, it was generated by default when I have added the new project, so I was thinking maybe it's the new microsoft best practice

Adaptors/S3/src/ArmoniK.Core.Adapters.S3.csproj
alouvion-aneo: I still don't know, I'm testing it with my bench project.

Common/src/Storage/IObjectStorage.cs
```diff
@@ -92,21 +92,6 @@ public static IObjectStorage CreateResourcesStorage(this IObjectStorageFactory f
...
-  ///   Update data if it already exists
-  /// </summary>
-  /// <param name="key">Key representing the object</param>
-  /// <param name="valueChunks">Chunks of data that will be stored in the Object Storage</param>
-  /// <param name="cancellationToken">Token used to cancel the execution of the method</param>
-  /// <returns>
-  ///   Task representing the asynchronous execution of the method
-  /// </returns>
-  /// <exception cref="ArmoniKException">there is 0 chunk</exception>
-  Task AddOrUpdateAsync(string                   key,
```
lemaitre-aneo, line 106: Why did you remove this method from the interface?

.github/workflows/build.yml
alouvion-aneo: I will do it afterwards : https://github.com/aneoconsulting/ArmoniK/issues/840

Adaptors/S3/src/ArmoniK.Core.Adapters.S3.csproj
alouvion-aneo: In the current implementation I finaly use it.
method comments are : Returns a <see cref="Stream"/> wrapping the contents of the given <see cref="ReadOnlyMemory{T}"/> of <see cref="byte"/> instance.
For the first implementation I think it is good to use it.
If we find better way to implement S3 read/write perfs in the bench part I will update it.

Common/src/Storage/IObjectStorage.cs
alouvion-aneo: seen with you

Adaptors/S3/src/ObjectStorage.cs
```diff
@@ -0,0 +1,249 @@
...
+      logger_.LogError("The key {key} was not found.",
+                       key);
+      throw new ObjectDataNotFoundException("Key not found");
+    }
+
+    // Get the data from the response stream
+    using var reader      = new StreamReader(response.ResponseStream);
+    var       fileContent = reader.ReadToEnd();
+
+    var valuesCount = int.Parse(fileContent!);
```
aneojgurhem, line 117: You should test if it is null and throw a proper exception if it is the case instead of the suppressing the nullable with `!`

Adaptors/S3/src/ObjectStorage.cs
```diff
@@ -0,0 +1,249 @@
...
+                                               .ConfigureAwait(false);
+    return deleteObjectsResponse.DeletedObjects.Count == valuesCount + 1;
+  }
+
+  /// <inheritdoc />
+  public IAsyncEnumerable<string> ListKeysAsync(CancellationToken cancellationToken = default)
+    => throw new NotImplementedException();
+}
+
+internal static class S3StorageHelper
```
aneojgurhem, line 190: This is an extension class for `AmazonS3Client ` so it should be called `AmazonS3ClientExt`

aneojgurhem (approved):

PR #224, opened 2023-01-27
feat(tools): script to download and apply format patch from the last workflow run on the current branch

aneojgurhem:
This PR adds a script to download and apply format patch from the last workflow run on the current branch.

### DIFF
--- a/tools/applyformatpatch.sh
+++ b/tools/applyformatpatch.sh
@@ -1,6 +1,9 @@
 #!/bin/sh
 
-rm -f patch.diff
+if test -e patch.diff ; then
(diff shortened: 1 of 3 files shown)


tools/applyformatpatch.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+
+rm -f patch.diff
```
jfonseca-aneo, line 3: You should check if this file exists first and  print a small message informing about its deletion

alouvion-aneo (changes requested):

alouvion-aneo (approved):

PR #228, opened 2023-01-30
Mongo Replicas and health check in terraform deployment

jfonseca-aneo:
Fix: https://github.com/aneoconsulting/ArmoniK/issues/849


terraform/modules/storage/database/mongo/outputs.tf
```diff
@@ -7,5 +7,6 @@ output "generated_env_vars" {
     "MongoDB__MaxConnectionPoolSize"         = "${var.mongodb_params.max_connection_pool_size}"
     "MongoDB__TableStorage__PollingDelayMin" = "${var.mongodb_params.min_polling_delay}"
     "MongoDB__TableStorage__PollingDelayMax" = "${var.mongodb_params.max_polling_delay}"
+    "MongoDB__DirectConnection"              = "true"
```
aneojgurhem, line 10: there is also a variable to set the mongo replicaset name. Can you add it there and put the name you put during deployment ?

jfonseca-aneo: done

aneojgurhem (approved):

PR #231, opened 2023-01-31
chunk items to delete by 1000

alouvion-aneo:
AWS S3 API has a limit of 1000 items to delete in his batch delete mode.
So if we want to delete more than 1000 items we have to chunk by 1000 or less.

### DIFF
--- a/Adaptors/S3/src/ObjectStorage.cs
+++ b/Adaptors/S3/src/ObjectStorage.cs
@@ -41,10 +41,10 @@ namespace ArmoniK.Core.Adapters.S3;
 
 public class ObjectStorage : IObjectStorage
 {
+  private const    int                    MaxAllowedKeysPerDelete = 1000;
   private readonly string                 bucketName_;
   private readonly ILogger<ObjectStorage> logger_;
   private readonly string                 objectStorageName_;
-  private readonly Options.S3             options_;
   private readonly AmazonS3Client         s3Client_;
 
   /// <summary>
@@ -61,7 +61,6 @@ public class ObjectStorage : IObjectStorage
   {
     s3Client_          = s3Client;
     objectStorageName_ = objectStorageName;
-    options_           = options;
     bucketName_        = options.BucketName;
     logger_            = logger;
   }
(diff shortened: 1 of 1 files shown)


Adaptors/S3/src/ObjectStorage.cs
```diff
@@ -169,17 +168,37 @@ await taskList.WhenAll()
...
-                                   };
+                                    });
 
-    var deleteObjectsResponse = await s3Client_.DeleteObjectsAsync(multiObjectDeleteRequest)
-                                               .ConfigureAwait(false);
-    return deleteObjectsResponse.DeletedObjects.Count == valuesCount + 1;
+    var deletedItem = 0;
+    try
+    {
+      var multiObjectDeleteRequest = new DeleteObjectsRequest
```
aneojgurhem, line 176: You need to create this object in the select.
You tried to use thread safe access for the add but not for the object list.

alouvion-aneo: indeed

alouvion-aneo: updated

Adaptors/S3/src/ObjectStorage.cs
```diff
@@ -169,17 +168,36 @@ await taskList.WhenAll()
...
+    {
+      await keyList.Chunk(MaxAllowedKeysPerDelete)
+                   .Select(async chunkedKeyList =>
+                           {
+                             var multiObjectDeleteRequest = new DeleteObjectsRequest
+                                                            {
+                                                              BucketName = bucketName_,
+                                                              Objects    = null,
+                                                            };
+                             multiObjectDeleteRequest.Objects = chunkedKeyList.ToList();
```
aneojgurhem, line 184: maybe you should be able to do something similar to:
```suggestion
                                                              Objects    =  {
                                                                  chunkedKeyList
                                                              },
                                                            };
```

alouvion-aneo: it is already the case... now...

aneojgurhem (approved):

PR #232, opened 2023-01-31
feat: Add helper for getting stats such as throughput

aneojgurhem:
This PR adds helpers to compute the throughput of the session and during the session (it counts task from other sessions if any). It also computes ratio between time spent on a task and processing time as well as the time spent between the end of the last dependency and the start of aggregation tasks.
Bench and HtcMock tests use this helper at their end.

### DIFF
--- a/Tests/Common/Client/src/GrpcChannelExt.cs
+++ b/Tests/Common/Client/src/GrpcChannelExt.cs
@@ -1,0 +1,200 @@
+// This file is part of the ArmoniK project
+//
+// Copyright (C) ANEO, 2021-$CURRENT_YEAR$. All rights reserved.
+//   W. Kirschenmann   <wkirschenmann@aneo.fr>
+//   J. Gurhem         <jgurhem@aneo.fr>
+//   D. Dubuc          <ddubuc@aneo.fr>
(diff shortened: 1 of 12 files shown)


alouvion-aneo (approved):

Tests/Common/Client/src/GrpcChannelExt.cs
```diff
@@ -0,0 +1,200 @@
...
+                                                         new ListTasksRequest.Types.Sort
+                                                         {
+                                                           Direction = ListTasksRequest.Types.OrderDirection.Asc,
+                                                           Field     = ListTasksRequest.Types.OrderByField.CreatedAt,
+                                                         })
+                                         .ConfigureAwait(false))
+    {
+      if (taskRaw.Status is TaskStatus.Completed or TaskStatus.Error)
+      {
+        var useRatio = (taskRaw.EndedAt - taskRaw.StartedAt).ToTimeSpan()
```
jfonseca-aneo, line 103: Dummy question: `taskRaw.StartetAT >= taskRaw.ReceivedAt`  right?

aneojgurhem: yes !

jfonseca-aneo (approved):

PR #244, opened 2023-02-07
feat: Add endpoint on Polling Agent to kill the current task if task is in cancelling status

aneojgurhem:
This PR adds an endpoint in the Agent to cancel the current task if the task is in cancelling state. It updates the taskData in the TaskHandler and eventually trigger the cancellation token if the task is in cancelling status.

fix https://github.com/aneoconsulting/ArmoniK/issues/510

### DIFF
--- a/Common/src/Pollster/TaskHandler.cs
+++ b/Common/src/Pollster/TaskHandler.cs
@@ -87,33 +87,35 @@ public class TaskHandler : IAsyncDisposable
                      Injection.Options.Pollster pollsterOptions,
                      CancellationTokenSource    cancellationTokenSource)
   {
-    sessionTable_            = sessionTable;
-    taskTable_               = taskTable;
-    resultTable_             = resultTable;
-    messageHandler_          = messageHandler;
-    taskProcessingChecker_   = taskProcessingChecker;
-    submitter_               = submitter;
-    dataPrefetcher_          = dataPrefetcher;
-    workerStreamHandler_     = workerStreamHandler;
-    activitySource_          = activitySource;
-    agentHandler_            = agentHandler;
-    logger_                  = logger;
-    cancellationTokenSource_ = cancellationTokenSource;
-    ownerPodId_              = ownerPodId;
-    ownerPodName_            = ownerPodName;
-    taskData_                = null;
-    sessionData_             = null;
+    sessionTable_          = sessionTable;
+    taskTable_             = taskTable;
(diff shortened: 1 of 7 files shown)


Common/src/Pollster/TaskHandler.cs
```diff
@@ -87,33 +87,35 @@ public class TaskHandler : IAsyncDisposable
...
+    workerConnectionCts_     = new CancellationTokenSource();
+    cancellationTokenSource_ = new CancellationTokenSource();
 
-    reg1_ = cancellationTokenSource_.Token.Register(() =>
-                                                    {
-                                                      logger_.LogWarning("Cancellation triggered, waiting {waitingTime} before cancelling task",
-                                                                         pollsterOptions.GraceDelay);
-                                                      workerConnectionCts_.CancelAfter(pollsterOptions.GraceDelay);
-                                                    });
+    reg1_ = cancellationTokenSource.Token.Register(() => cancellationTokenSource_.Cancel());
```
ngruelaneo, line 111: cela ne serait il pas plutot?

`reg1_ = cancellationTokenSource_.Token.Register(()`

Common/src/Pollster/TaskHandler.cs
```diff
@@ -131,10 +133,32 @@ await messageHandler_.DisposeAsync()
     reg1_.Unregister();
     await reg1_.DisposeAsync()
                .ConfigureAwait(false);
+    cancellationTokenSource_.Dispose();
     workerConnectionCts_.Dispose();
     agent_?.Dispose();
   }
 
+  /// <summary>
+  ///   Refresh task metadata and cancel execution if current task should be cancelled
```
ngruelaneo, line 142:
```suggestion
  ///   Refresh task metadata and stop execution if current task should be cancelled
```

Common/src/Pollster/TaskHandler.cs
aneojgurhem: non c'est intentionnel. Je fais une copie du token sinon ça cancel TOUT. Mais VRAIMENT TOUT.

Common/src/Pollster/TaskHandler.cs
aneojgurhem: ok je change

ngruelaneo (approved):

ngruelaneo (approved):

PR #245, opened 2023-02-08
feat(api): Implement count tasks methods from new api version

aneojgurhem:
This PR implements the new APIs introduced in https://github.com/aneoconsulting/ArmoniK.Api/pull/142

### DIFF
--- a/Common/src/gRPC/Services/GrpcApplicationsService.cs
+++ b/Common/src/gRPC/Services/GrpcApplicationsService.cs
@@ -22,12 +22,15 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+using System;
 using System.Linq;
 using System.Threading.Tasks;
 
+using ArmoniK.Api.gRPC.V1;
 using ArmoniK.Api.gRPC.V1.Applications;
 using ArmoniK.Core.Common.Auth.Authentication;
 using ArmoniK.Core.Common.Auth.Authorization;
+using ArmoniK.Core.Common.Exceptions;
 using ArmoniK.Core.Common.Storage;
 
 using Grpc.Core;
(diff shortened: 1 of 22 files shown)


Common/src/gRPC/Services/GrpcApplicationsService.cs
```diff
@@ -79,4 +82,41 @@ public class GrpcApplicationsService : Applications.ApplicationsBase
...
+      throw new RpcException(new Status(StatusCode.Internal,
+                                        "Internal ArmoniK Exception, see application logs"));
+    }
+    catch (Exception e)
+    {
+      logger_.LogWarning(e,
+                         "Error while counting tasks by applications");
+      throw new RpcException(new Status(StatusCode.Unknown,
+                                        "Unknown Exception, see application logs"));
+    }
```
lemaitre-aneo, line 120: Don't you need to catch some exceptions to make them user errors, instead of internal errors?

Common/tests/Auth/AuthenticationIntegrationTest.cs
```diff
@@ -110,40 +111,43 @@ await helper_.StopServer()
...
+  private static readonly ListApplicationsRequest                        ListApplicationsRequest;
+  private static readonly CancelTasksRequest                             CancelTasksRequest;
+  private static readonly ListResultsRequest                             ListResultsRequest;
+  private static readonly GetCurrentUserRequest                          GetCurrentUserRequest;
+  private static readonly EventSubscriptionRequest                       EventSubscriptionRequest;
+  private static readonly ListPartitionsRequest                          ListPartitionsRequest;
+  private static readonly GetPartitionRequest                            GetPartitionRequest;
+  private static readonly CountTasksByStatusRequest                      CountTasksByStatusRequestApplications;
+  private static readonly Api.gRPC.V1.Sessions.CountTasksByStatusRequest CountTasksByStatusRequestSessions;
+  private static readonly Api.gRPC.V1.Tasks.CountTasksByStatusRequest    CountTasksByStatusRequestTasks;
```
lemaitre-aneo, line 150: Couldn't you use a namespace alias to avoid the really long FQN?

Common/src/gRPC/Services/GrpcApplicationsService.cs
aneojgurhem: The underlyning method does not throw exceptions. It will just return empty list most of the time.

Common/tests/Auth/AuthenticationIntegrationTest.cs
aneojgurhem: I use it for the first application. The namespace proposition is not really great

lemaitre-aneo: No, I mean something like:
```c#
using grpc = Api.gRPC.V1;
```

In retrospective, I am not sure this would gain much.

lemaitre-aneo: > I use it for the first count..... The namespace proposition is not really great

You should avoid using the non qualified name for the Application one to make it consistent, though.

aneojgurhem: I'm not really convinced for this one.
Iwould have prefered something like

```csharp
using ApplicationCountTasksByStatusRequest Api.gRPC.V1.Applications.CountTasksByStatusRequest
```

aneojgurhem: > > I use it for the first count..... The namespace proposition is not really great
> 
> You should avoid using the non qualified name for the Application one to make it consistent, though.

well, jetbrains is not ok with that

Common/src/gRPC/Services/GrpcApplicationsService.cs
lemaitre-aneo: fair enough

Common/tests/Auth/AuthenticationIntegrationTest.cs
lemaitre-aneo: That's a bummer...

lemaitre-aneo (approved):

PR #249, opened 2023-02-10
ci: Change release process so that we can create tags and release through github

aneojgurhem:
Change the release creation process so that we can create tags and release through github. This is a step to uniformize release processes accross our repositories.

The release workflow for new merged code on main would be:
- Create a release from github interface
- Choose the proper next version tag according to the features (feat:) present in the history from the last tag.
- Publish the release
- If the tag is valid, docker images with the new tag are released
- If the tag is __not__ valid, tag is deleted and released is put in draft.

### DIFF
--- a/.github/workflows/make-release.yml
+++ b/.github/workflows/make-release.yml
@@ -24,7 +24,6 @@ jobs:
     - name: Remove tag locally
     # This is done so that codacy/git-version can compute the version of the images deployed on docker hub
       run: |
-        git checkout $(git rev-parse HEAD)
         git tag -d ${{ github.ref_name }}
 
     - name: Compute branch for codacy
--- a/.github/workflows/make-release.yml
+++ b/.github/workflows/make-release.yml
@@ -1,16 +1,18 @@
 name: Release
 
+permissions:
+  contents: write
+
 on:
   push:
-    branches:
-      - release
-    tags-ignore: [ '**' ]
+    tags:
+      - "[0-9]+.[0-9]+.[0-9]+"
 
 jobs:
   versionning:
     runs-on: ubuntu-latest
     outputs:
-      version: ${{ steps.pre-rel.outputs.version }}
+      version: ${{ steps.snapshot.outputs.version }}
       release: ${{ steps.release.outputs.version }}
     steps:
     - name: Checkout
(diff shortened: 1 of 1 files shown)


jfonseca-aneo (approved):

lemaitre-aneo:
What would be the workflow to do a release? Would it make sense for this to be a manual workflow?

.github/workflows/make-release.yml
```diff
@@ -19,26 +21,46 @@ jobs:
         ref: ${{ github.ref }}
         fetch-depth: 0
 
-    - name: Generate Pre-Release Version
-      id: pre-rel
+    - name: Remove tag locally
+    # This is done so that codacy/git-version can compute the version of the images deployed on docker hub
+      run: |
+        git checkout $(git rev-parse HEAD)
```
ngruelaneo, line 27: pourquoi as tu besoin de cela?

aneojgurhem:
> What would be the workflow to do a release? Would it make sense for this to be a manual workflow?

The release workflow for new merged code on main would be:
- Create a release from github interface
- Choose the proper next version tag according to the features (feat:) present in the history from the last tag.
- Publish the release
- If the tag is valid, docker images with the new tag are released
- If the tag is __not__ valid, tag is deleted and released is put in draft.

.github/workflows/make-release.yml
aneojgurhem: parce que je pensais que tu ne pouvais pas supprimer le tag quand tu avais fait un checkout dessus

.github/workflows/make-release.yml
```diff
@@ -19,26 +21,45 @@ jobs:
         ref: ${{ github.ref }}
```
lemaitre-aneo, line 21: Do you really need to keep the `ref` argument?

lemaitre-aneo (approved):

.github/workflows/make-release.yml
ngruelaneo: je pense que cela ne pose pas de probleme mais ok

.github/workflows/make-release.yml
aneojgurhem: Yes ! Sometimes, the checkout is not properly done. It happened that I did not get the modifications from my PR into the resulting package...

PR #250, opened 2023-02-13
chore: remove names from headers

aneojgurhem:
This PR removes the names from the headers in C# files. They will be added in another files later on.


alouvion-aneo (approved):

.editorconfig
```diff
@@ -65,7 +65,7 @@ dotnet_style_qualification_for_field = false:suggestion
...
+file_header_template = This file is part of the ArmoniK project\n\nCopyright (C) ANEO, 2021-$CURRENT_YEAR$. All rights reserved.\n\nThis program is free software: you can redistribute it and/or modify\nit under the terms of the GNU Affero General Public License as published\nby the Free Software Foundation, either version 3 of the License, or\n(at your option) any later version.\n\nThis program is distributed in the hope that it will be useful,\nbut WITHOUT ANY WARRANTY, without even the implied warranty of\nMERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the\nGNU Affero General Public License for more details.\n\nYou should have received a copy of the GNU Affero General Public License\nalong with this program.  If not, see <http://www.gnu.org/licenses/>.
```
lemaitre-aneo, line 68: I would take the opportunity to remove the `$CURRENT_YEAR$`, and replace it with `2023`. We will have to change files anyway next year, so we can change this file year in addition.

aneojgurhem: But this file does not have to change when we update year

lemaitre-aneo: Yes, but in most of our Dev environment, the variable replacement is not performed. Changing it here would solve the issue altogether.

aneojgurhem: It is not performed because code cleanup is poorly configured. Putting a fixed value here does not mean that we will change it regularly.

ngruelaneo: Indeed I think it is better to remove the date completely

aneojgurhem: The date has to stay. It is a requirement for the copyright

lemaitre-aneo: > It is not performed because code cleanup is poorly configured. Putting a fixed value here does not mean that we will change it regularly.

In theory, I agree that it is only a configuration problem, but in practice, it is something that nobody except you know how to setup, and thus, you are the only one where the replacement is done.
My point is: the change is once a year, and must be done on all files. So changing it in one extra file is no extra burden.

lemaitre-aneo: (this is not blocking the PR, so if you want to merge, I have approved it)

PR #256, opened 2023-02-15
feat: implement multiple sort field when list applications

aneojgurhem:
This PR implements multiple sort field when list applications for GUI.

fix: https://github.com/aneoconsulting/ArmoniK/issues/896

### DIFF
--- a/Adaptors/MongoDB/src/IMongoQueryableExt.cs
+++ b/Adaptors/MongoDB/src/IMongoQueryableExt.cs
@@ -44,9 +44,9 @@ internal static class IMongoQueryableExt
     }
   }
 
-  public static IOrderedMongoQueryable<T> OrderByList<T>(this IMongoQueryable<T>            queryable,
-                                                         List<Expression<Func<T, object?>>> orderFields,
-                                                         bool                               ascOrder = true)
+  public static IOrderedMongoQueryable<T> OrderByList<T>(this IMongoQueryable<T>                   queryable,
+                                                         ICollection<Expression<Func<T, object?>>> orderFields,
+                                                         bool                                      ascOrder = true)
   {
     if (ascOrder)
     {
--- a/Common/src/Utils/IQueryableExt.cs
+++ b/Common/src/Utils/IQueryableExt.cs
@@ -19,34 +19,14 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Linq.Expressions;
-using System.Runtime.CompilerServices;
-using System.Threading;
 
-using MongoDB.Driver.Linq;
+namespace ArmoniK.Core.Common.Utils;
 
-namespace ArmoniK.Core.Adapters.MongoDB;
-
-internal static class IMongoQueryableExt
+public static class IQueryableExt
 {
-  public static async IAsyncEnumerable<T> ToAsyncEnumerable<T>(this                     IMongoQueryable<T> queryable,
-                                                               [EnumeratorCancellation] CancellationToken  cancellationToken = default)
-  {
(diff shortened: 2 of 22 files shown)


Adaptors/MongoDB/src/IMongoQueryableExt.cs
```diff
@@ -40,4 +43,36 @@ internal static class IMongoQueryableExt
       }
     }
   }
+
+  public static IOrderedMongoQueryable<T> OrderByList<T>(this IMongoQueryable<T>            queryable,
+                                                         List<Expression<Func<T, object?>>> orderFields,
```
lemaitre-aneo, line 48:
```suggestion
                                                         ICollection<Expression<Func<T, object?>>> orderFields,
```
I think there is no need for a List specifically. You could at least take a `ICollection`, and maybe a `IEnumerable` even.

Adaptors/MongoDB/src/IMongoQueryableExt.cs
```diff
@@ -40,4 +43,36 @@ internal static class IMongoQueryableExt
...
+                                                         bool                               ascOrder = true)
```
lemaitre-aneo, line 49: Order should most likely be per "column", and not necessarily the same for all "columns".

Common/src/Utils/IQueryableExt.cs
```diff
@@ -0,0 +1,58 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Linq.Expressions;
+
+namespace ArmoniK.Core.Common.Utils;
+
+public static class IQueryableExt
+{
+  public static IOrderedQueryable<T> OrderByList<T>(this IQueryable<T>                 queryable,
```
lemaitre-aneo, line 27: Do you really have to implement it twice?

Adaptors/MongoDB/src/IMongoQueryableExt.cs
aneojgurhem: ICollection is effectively good. We will have a multiple iteration problem with IEnumerable due to the first and skip.

Common/src/Utils/IQueryableExt.cs
aneojgurhem: due to https://github.com/aneoconsulting/ArmoniK.Core/blob/13805d638c8c0162c88070c6649ae53aba8934dc/Adaptors/MongoDB/src/TaskTable.cs#L473
that we created, we cannot change the type of the output of the orderby. For now, I do not have a better solution

Adaptors/MongoDB/src/IMongoQueryableExt.cs
aneojgurhem: It is not needed and simplifies code for now

lemaitre-aneo (commented):
It would be nice to test if the sort actually works.

Common/tests/TestBase/TaskTableTestBase.cs
```diff
@@ -1250,6 +1254,151 @@ public async Task ListApplicationFromTasksShouldSucceed()
...
+                                     .ConfigureAwait(false);
+
+      var listTasksResponseTaskData = listTasks.applications.ToList();
+      foreach (var task in listTasksResponseTaskData)
+      {
+        Console.WriteLine(task);
+      }
+
+      // app2 before app1 because desc ordering in request
+      // app2 before app3 because desc and ordering by version then by service
```
lemaitre-aneo, line 1391: From the comment, we cannot say which app1 or app3 should be first.

lemaitre-aneo (approved):

PR #262, opened 2023-02-24
feat: Add a service for dependency resolution

aneojgurhem:
This PR adds a service for a better management of data dependencies for tasks.

- [x] Implement dependency checker service
- [x] Agent register tasks that depends on the results submitted by the running task and put them in dependency resolver queue we task completes successfully
- [x] Tests for DependencyResolver
- [x] Tests for ITaskTable new method
- [x] Tests for new fonctionnality in agent
- [ ] Test new task queuing in Submitter FinalizeTaskCreation method
- [x] Doc for public members
- [x] Add tests for new IResultTable methods

fix: https://github.com/aneoconsulting/ArmoniK/issues/858


Common/src/gRPC/Services/Agent.cs
```diff
@@ -428,8 +453,9 @@ await foreach (var request in requestStream.ReadAllAsync(cancellationToken)
                                           ("taskId", taskData_.TaskId),
                                           ("sessionId", sessionData_.SessionId));
 
-    Task? completionTask = null;
-    var   fsmResult      = new ProcessReplyResultStateMachine(logger_);
+    Task? completionTask       = null;
+    Task? findDependenciesTask = null;
```
lemaitre-aneo, line 457: You can initialize those with `Task.CompletedTask` or `Task.FromCanceled` to avoid the nullable.

terraform/modules/compute_plane/locals.tf
```diff
@@ -10,7 +10,8 @@ locals {
     "InitWorker__WorkerCheckRetries=${var.polling_agent.worker_check_retries}",
     "InitWorker__WorkerCheckDelay=${var.polling_agent.worker_check_retries}",
     "Zipkin__Uri=${var.zipkin_uri}",
-    "Amqp__PartitionId=TestPartition${local.partition_chooser}"
+    "Amqp__PartitionId=TestPartition${local.partition_chooser}",
+    "DependencyResolver__UnresolvedDependenciesQueue=${var.unresolved_dependencies_queue}"
```
lemaitre-aneo, line 14: Add a final comma

Adaptors/Memory/src/ResultTable.cs
```diff
@@ -49,14 +50,15 @@ public class ResultTable : IResultTable
   }
 
   /// <inheritdoc />
-  public Task<IEnumerable<ResultStatusCount>> AreResultsAvailableAsync(string              sessionId,
-                                                                       IEnumerable<string> keys,
-                                                                       CancellationToken   cancellationToken = default)
+  public Task<IList<ResultStatusCount>> AreResultsAvailableAsync(string              sessionId,
```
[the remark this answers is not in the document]
aneojgurhem: We need to iterate multiple times over this result.
I tried with collections but it did not work as smoothly as I would have liked. I'll try again.

Common/src/gRPC/Services/Agent.cs
```diff
@@ -44,37 +44,53 @@ namespace ArmoniK.Core.Common.gRPC.Services;
 /// </summary>
 public class Agent : IAgent
 {
+  private sealed record TaskIdPriority
+  {
+    public string TaskId   { get; init; } = string.Empty;
```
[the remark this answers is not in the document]
aneojgurhem: So we need to uniformize. I prefer the empty version because we can assert they are the same object

Common/src/DependencyResolver/DependencyResolver.cs
```diff
@@ -0,0 +1,181 @@
...
+                                                           ("messageHandler", message.MessageId),
+                                                           ("taskId", message.TaskId));
+
+          var taskData = await taskTable_.ReadTaskAsync(message.TaskId,
+                                                        stoppingToken)
+                                         .ConfigureAwait(false);
+
+          switch (taskData.Status)
+          {
+            case TaskStatus.Creating:
```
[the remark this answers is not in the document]
aneojgurhem: I was thinking the same. It is not strictly necessary but I would be clearer.
A kind of WaitingForDataDependencies ?

Common/src/gRPC/Services/Agent.cs
```diff
@@ -474,6 +500,19 @@ await foreach (var request in requestStream.ReadAllAsync(cancellationToken)
                                                           .ConfigureAwait(false);
                                         },
                                         cancellationToken);
+              findDependenciesTask = Task.Run(async () =>
+                                              {
+                                                var taskFound = await taskTable_.FindTasksAsync(data => data.DataDependencies.Contains(request.Init.Key),
```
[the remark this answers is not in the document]
aneojgurhem: I wanted to keep tasks in the queue, so I need to get the task Ids.
I don't think that this request is well optmized since we search in a list from TaskData objects.
It was simpler this way. We can do so in a second time. Adding the reverse dependency list in the Results would have made more modifications.

terraform/modules/compute_plane/locals.tf
aneojgurhem: We will need to check terraform files formatting and style

Common/src/gRPC/Services/Agent.cs
aneojgurhem: We will need to properly activate nullables in Core. There are issues with it. Sometimes there are hint saying that nullable are not activated.

Adaptors/Memory/src/ResultTable.cs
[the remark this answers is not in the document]
aneojgurhem: I put an IEnumerable back

Common/src/gRPC/Services/Agent.cs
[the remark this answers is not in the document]
lemaitre-aneo: I think we mostly use `""`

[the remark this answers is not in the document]
aneojgurhem: It depends where. Since we did not clarify the usage of `""` or `string.Empty` we use both

terraform/modules/dependency_checker/versions.tf
```diff
@@ -0,0 +1,8 @@
+terraform {
+  required_providers {
+    docker = {
+      source  = "kreuzwerker/docker"
+      version = ">= 2.25.0"
```
jfonseca-aneo, line 5: This has been updated to ` 3.0.1`

Common/tests/DependencyResolverTest.cs
```diff
@@ -0,0 +1,587 @@
...
+  private const string AvailableData1                        = nameof(AvailableData1);
+  private const string AvailableData2                        = nameof(AvailableData2);
+  private const string TaskCreatingNoDependencies            = nameof(TaskCreatingNoDependencies);
+  private const string TaskSubmitted                         = nameof(TaskSubmitted);
+  private const string TaskCancelling                        = nameof(TaskCancelling);
+  private const string TaskCancelled                         = nameof(TaskCancelled);
+
+  public class MyMessageHandler : IQueueMessageHandler
+  {
+    public MyMessageHandler(CancellationToken  cancellationToken,
```
jfonseca-aneo, line 69: I'd isolate this in a helper

Common/tests/DependencyResolverTest.cs
```diff
@@ -0,0 +1,587 @@
...
+                                          Guid.NewGuid()
+                                              .ToString(),
+                                          m,
+                                          QueueMessageStatus.Running,
+                                          DateTime.UtcNow);
+      }
+    }
+  }
+
+  public class MyPushQueueStorage : IPushQueueStorage
```
jfonseca-aneo, line 125: and this as well

Common/tests/DependencyResolverTest.cs
```diff
@@ -0,0 +1,587 @@
...
+                                                          "sad task")),
+                                })
+                   .ConfigureAwait(false);
+
+    return provider;
+  }
+
+
+  [Test]
+  public async Task Init()
```
jfonseca-aneo, line 480: This looks more like your `[SetUp]`

Common/tests/DependencyResolverTest.cs
```diff
@@ -0,0 +1,587 @@
...
+                                                           logger,
+                                                           CancellationToken.None)
+                                    .ConfigureAwait(false);
+  }
+
+
+  [Test]
+  public async Task CheckForDependenciesShouldSucceed()
+  {
+    using var provider = await Populate()
```
jfonseca-aneo, line 523: So you can avoid calling this block

Common/src/DependencyResolver/DependencyResolver.cs
```diff
@@ -0,0 +1,181 @@
...
+
+              var dependenciesStatus = await TaskLifeCycleHelper.CheckTaskDependencies(taskData,
+                                                                                       resultTable_,
+                                                                                       logger_,
+                                                                                       stoppingToken)
+                                                                .ConfigureAwait(false);
+              logger_.LogInformation("task dependencies : {resolved}",
```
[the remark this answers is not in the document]
aneojgurhem: The taskId is included as a scope in every log of the dependency resolver

Common/tests/Helpers/SimpleResultTable.cs
```diff
@@ -51,11 +51,11 @@ public Task Init(CancellationToken cancellationToken)
   public Task<IEnumerable<ResultStatusCount>> AreResultsAvailableAsync(string              sessionId,
                                                                        IEnumerable<string> keys,
                                                                        CancellationToken   cancellationToken = default)
-    => Task.FromResult<IEnumerable<ResultStatusCount>>(new ResultStatusCount[]
-                                                       {
-                                                         new(ResultStatus.Completed,
-                                                             1),
-                                                       });
+    => Task.FromResult(new List<ResultStatusCount>
```
jfonseca-aneo, line 54: What is the difference in constructing the output this way?

Common/src/DependencyResolver/DependencyResolver.cs
```diff
@@ -0,0 +1,181 @@
...
+    logger_           = logger;
+  }
+
+  /// <inheritdoc />
+  public Task<HealthCheckResult> Check(HealthCheckTag tag)
+    => Task.FromResult(HealthCheckResult.Healthy());
+
+  /// <inheritdoc />
+  public Task Init(CancellationToken cancellationToken)
+    => Task.CompletedTask;
```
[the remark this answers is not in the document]
aneojgurhem: done !

Common/tests/DependencyResolverTest.cs
aneojgurhem: I did not want to struggle with the provider that should be nullable in this case

Common/tests/Helpers/SimpleResultTable.cs
aneojgurhem: I changed this method so that it returns a `Ilist` then put a `IEnumerable` again and did not put back the previous implementation.
It avoid a cast from the `Task.FromResult` but `AsEnumerable` may do so too. Actually, it does not change much. I find it more readable now

Common/src/DependencyResolver/DependencyResolver.cs
```diff
@@ -0,0 +1,256 @@
...
+  public async Task Init(CancellationToken cancellationToken)
+  {
+    await pushQueueStorage_.Init(cancellationToken)
+                           .ConfigureAwait(false);
+    await pullQueueStorage_.Init(cancellationToken)
+                           .ConfigureAwait(false);
+    await resultTable_.Init(cancellationToken)
+                      .ConfigureAwait(false);
+    await taskTable_.Init(cancellationToken)
+                    .ConfigureAwait(false);
```
lemaitre-aneo, line 127: Would it make sense to make them in parallel?
```suggestion
    await Task.WhenAll(new []
                       {
                         pushQueueStorage_.Init(cancellationToken),
                         pullQueueStorage_.Init(cancellationToken),
                         resultTable_.Init(cancellationToken),
                         taskTable_.Init(cancellationToken),
                       })
              .ConfigureAwait(false);
```

Common/src/Pollster/AgentHandler.cs
```diff
@@ -43,31 +43,47 @@ namespace ArmoniK.Core.Common.Pollster;
...
   ///   Initializes a new instance
   /// </summary>
   /// <param name="loggerInit">Logger initializer used to configure the loggers needed by the worker</param>
   /// <param name="computePlaneOptions">Options needed for the creation of the servers</param>
+  /// <param name="dependencyResolverOptions">Configuration for DependencyResolver</param>
   /// <param name="submitter">Interface to manage tasks</param>
   /// <param name="objectStorageFactory">Interface class to create object storage</param>
+  /// <param name="pushQueueStorage">Interface to put tasks in the queue</param>
+  /// <param name="taskTable">Interface to manage task states</param>
+  /// <param name="resultTable">Interface to manage result states</param>
```
lemaitre-aneo, line 67: That begins to be too many params, just for the sake of being passed to the Agent. Couldn't we rely on the dependency injection to build the Agent?

(Could be done in another PR)

Common/src/gRPC/Services/Submitter.cs
```diff
@@ -115,22 +118,66 @@ public class Submitter : ISubmitter
...
+      {
+        await resultTable_.AddTaskDependency(sessionId,
+                                             dependencies,
+                                             new List<string>
+                                             {
+                                               request.Id,
+                                             },
+                                             cancellationToken)
+                          .ConfigureAwait(false);
+        dependencies = await resultTable_.GetResults(sessionId,
```
lemaitre-aneo, line 155: @lemaitre-aneo add a comment to explain why the double get is needed.

Common/src/Pollster/AgentHandler.cs
aneojgurhem: With the way it is created, I don't know if we can do it easily with DI. Moreover, we need a new `Agent` for every task. The `AgentHandler` is already created with DI.

Common/src/DependencyResolver/DependencyResolver.cs
aneojgurhem: Well, it was not done in parallel in the other Init so I did the same. It makes sense but I'm not sure if it will break something.

jfonseca-aneo (dismissed):

jfonseca-aneo (approved):

PR #277, opened 2023-03-03
refactor: Miscellaneous improvement on terraform

lemaitre-aneo:
Some improvements to the terraform


terraform/modules/storage/object/minio/outputs.tf
```diff
@@ -7,4 +7,9 @@ output "generated_env_vars" {
     "S3__Password"              = var.password
     "S3__MustForcePathStyle"    = true
   })
-}
\ No newline at end of file
+}
+
+output "volumes" {
+  description = "Volumes that agents must mount to access the object storage"
```
aneojgurhem, line 13: not only agents

lemaitre-aneo: Arguably, the submitter is also an agent. What naming would you suggest?

aneojgurhem: agents and submitters

aneojgurhem (approved):

PR #280, opened 2023-03-03
fix: use an ObjectPool to store Sessions instead of creating one at each push in AMQP adapter

aneojgurhem:
Introduce an ObjectPool to store Sessions instead of creating one at each push in AMQP adapter. It should solve issues when timeout appears when trying to close session after sending messages into the queue

### DIFF
--- a/Adaptors/Amqp/src/PushQueueStorage.cs
+++ b/Adaptors/Amqp/src/PushQueueStorage.cs
@@ -29,23 +29,50 @@ using ArmoniK.Core.Common.Exceptions;
 using ArmoniK.Core.Common.Storage;
 
 using Microsoft.Extensions.Logging;
-
-using TimeoutException = System.TimeoutException;
+using Microsoft.Extensions.ObjectPool;
 
 namespace ArmoniK.Core.Adapters.Amqp;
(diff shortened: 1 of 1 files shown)


jfonseca-aneo (approved):
Can you explain why does it improve the perfs in https://github.com/aneoconsulting/ArmoniK.Core/pull/262 ? Are you doing significantly more pushes there?

aneojgurhem:
> Can you explain why does it improve the perfs in #262 ? Are you doing significantly more pushes there?

Yes! We insert the taskId in the dependency resolver queue. The agent has to insert the ids from the tasks that depend from the submitted results. We observed that the request to the db and the insertion in the queue take a significant time. I added this Pool and it improved the performances

lemaitre-aneo:
Could it introduce more issues with the number of connections to the queue? For now, on AWS we are limited to 2000 connections.

aneojgurhem:
> Could it introduce more issues with the number of connections to the queue? For now, on AWS we are limited to 2000 connections.

It creates multiple sessions from one connection

lemaitre-aneo (approved):

PR #289, opened 2023-03-15
perf: Parallel result retrieving in bench test

aneojgurhem:
Improve performances of results retrieving by getting multiple results at the same time.

### DIFF
--- a/Tests/Bench/Client/src/Program.cs
+++ b/Tests/Bench/Client/src/Program.cs
@@ -289,6 +289,12 @@ internal static class Program
                                                             CancellationToken.None)
                                             .Result;
 
+                         // A good a way to process results would be to process them individually as soon as they are
+                         // retrieved. They may be stored in a ConcurrentBag or a ConcurrentDictionary but you need to
+                         // be careful to not overload your memory. If you need to retrieve a lot of results to apply
+                         // post-processing on, consider doing so with sub-tasking so that the client-side application
+                         // has to do less work.
+
                          if (result.Length != benchOptions.ResultSize * 1024)
                          {
                            logger.LogInformation("Received length {received}, expected length {expected}",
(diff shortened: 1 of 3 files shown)


Tests/Bench/Client/src/Program.cs
```diff
@@ -239,24 +266,62 @@ private static async Task Main()
 
     var resultsAvailable = Stopwatch.GetTimestamp();
 
-    foreach (var resultId in results)
+    var countRes = 0;
+
+    results.AsParallel()
```
lemaitre-aneo, line 271: Do we want spawning threads and do blocking IO, or using async/await with a semaphore?

Tests/Bench/Client/src/Program.cs
```diff
@@ -239,24 +266,62 @@ private static async Task Main()
...
+                       {
+                         var resultRequest = new ResultRequest
+                                             {
+                                               ResultId = resultId,
+                                               Session  = createSessionReply.SessionId,
+                                             };
+
+                         var client = new Submitter.SubmitterClient(localChannel);
+
+                         var result = client.GetResultAsync(resultRequest,
```
jfonseca-aneo, line 288: Even if we are not doing anything with the results afterwards, I think they should be put in a thread safe storage object one they are retrieved, isn't it?

aneojgurhem: I don't want to store them. They can be really large and they will only stop us from running very large tests

Tests/Bench/Client/src/Program.cs
aneojgurhem: Not today

Tests/Bench/Client/src/Program.cs
jfonseca-aneo: Ok, I'd then add a comment on the "right way" to store them, in case a future user might use this as en example  of how to recuperate the results.

jfonseca-aneo (approved):

PR #291, opened 2023-03-15
style: rewrite terraform files to a canonical format

melflitty:
Formatting terraform configuration files

### DIFF
--- a/terraform/modules/storage/database/mongo/locals.tf
+++ b/terraform/modules/storage/database/mongo/locals.tf
@@ -1,3 +1,3 @@
 locals {
-    test-cmd = "exec test $$( mongosh --quiet --eval 'rs.status().ok' ) -eq 1"
+  test-cmd = "exec test $$( mongosh --quiet --eval 'rs.status().ok' ) -eq 1"
 }
--- a/terraform/modules/compute_plane/main.tf
+++ b/terraform/modules/compute_plane/main.tf
@@ -104,5 +104,5 @@ resource "docker_container" "polling_agent" {
     retries      = 5
   }
 
-  depends_on = [ docker_container.worker ]
+  depends_on = [docker_container.worker]
 }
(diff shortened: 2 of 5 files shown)


terraform/modules/storage/database/mongo/locals.tf
```diff
@@ -1,3 +1,3 @@
 locals {
-    test-cmd = "exec test $$( mongosh --quiet --eval 'rs.status().ok' ) -eq 1"
+  test-cmd = "exec test $$( mongosh --quiet --eval 'rs.status().ok' ) -eq 1"
 }
```
lemaitre-aneo, line 3: Canonical format should include a an empty line at the end of the file

aneojgurhem: `terraform fmt` does not add it

lemaitre-aneo: I know, but I think it should still be part of _our_ conventions (like LF line returns). Maybe we should apply dos2unix on all files.

aneojgurhem: So we also need a way to ensure it. The purpose of this formatting is to validate terraform formatting in https://github.com/aneoconsulting/ArmoniK.Core/pull/287

lemaitre-aneo: I would say, even though it is not exactly required, it can be nice to take the opportunity of this PR to run at least dos2unix on all the text files, because we know that's something we will want anyway in the future. We could then investigate those tools in another PR.

```sh
find . -type f | xargs dos2unix
```

aneojgurhem: We'll need to exclude .git files...

aneojgurhem: I tried on terraform files and they are already unix files

aneojgurhem: We could add more check in the formatting pipeline

aneojgurhem (approved):

PR #296, opened 2023-03-21
ci: checkout on the right reference in terraform format checker

aneojgurhem:
Fix the checkout to the branch in Terraform format check.

### DIFF
--- a/.github/workflows/code-formatting.yml
+++ b/.github/workflows/code-formatting.yml
@@ -62,7 +62,7 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
         with:
-          ref: ${{ github.head_ref }}
+          ref: ${{ github.ref }}
 
       - name: Check Format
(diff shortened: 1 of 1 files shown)


.github/workflows/code-formatting.yml
```diff
@@ -62,7 +62,7 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
         with:
-          ref: ${{ github.head_ref }}
+          ref: ${{ github.ref }}
```
lemaitre-aneo, line 65: I might be late to the party, but you could even just remove the ref argument.

aneojgurhem: We need this reference. I had some issues with the checkout made to a branch I did not expect

lemaitre-aneo: As far as I remember, without the option, it checks out the expected merge version, which is what you need to verify. I assume the problem you had was about tagging.

PR #306, opened 2023-03-30
Fix issue with AWSSDK.SecurityToken

alouvion-aneo:
Fix issue with missing package when SecurityToken is used on S3


aneojgurhem (approved):

alouvion-aneo:
closed because I need it in 0.12.3 first then merge it to main

PR #308, opened 2023-04-03
feat: complete terraform monitoring modules to Core deployment

melflitty:
Fix: https://github.com/aneoconsulting/ArmoniK/issues/943


terraform/modules/monitoring/grafana/inputs.tf
```diff
@@ -0,0 +1,14 @@
+variable "image" {
+  type = string
+}
+
+variable "network" {
+  type = string
+
```
aneojgurhem, line 7: I don't remember if we should put defaut values here ! Can you look in the other modules and do the same ?

terraform/prometheus/prometheus.yaml
```diff
@@ -0,0 +1,14 @@
...
+  scrape_interval:     15s
+  evaluation_interval: 15s
+
+scrape_configs:
+  - job_name: 'prometheus'
+    static_configs:
+      - targets: [ 'localhost:9090' ]
+  - job_name: 'tempo'
+    static_configs:
+      - targets: [ 'tempo:3200' ]
```
aneojgurhem, line 11: Is there a variable that could change this value in the tempo module or the metrics exporter ? If there is we should paramterize the two values.

melflitty: You mean to parametrize the port number?(if it can change)

terraform/prometheus/prometheus.yaml
```diff
@@ -0,0 +1,14 @@
...
+  - job_name: 'armonik.control.metrics'
+    static_configs:
+      - targets: [ 'armonik.control.metrics:1080' ]
```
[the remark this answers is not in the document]
melflitty: Done

terraform/modules/monitoring/grafana/inputs.tf
aneojgurhem: Ok for me

terraform/prometheus/prometheus.yaml
aneojgurhem: I mean : If we already have an option that allows to change the port, we need to make sure that we use this parameter. If we do not have something like that, it is ok for me if we let it like that

aneojgurhem (approved):

PR #312, opened 2023-04-07
feat: control plane creates results

aneojgurhem:
This implements rpc method to create results metadata from control plane instead of generating it client side.
It also changes the task submission to expect created results instead of creating them implicitly.

fix https://github.com/aneoconsulting/ArmoniK/issues/439


dbrasseur-aneo (changes requested):

Common/src/Storage/IResultTable.cs
```diff
@@ -219,6 +219,19 @@ public interface IResultTable : IInitializable
...
+  ///   Set Task that should produce the result
+  /// </summary>
+  /// <param name="sessionId"></param>
+  /// <param name="requests">Results to update with the associated task id</param>
+  /// <param name="cancellationToken">Token used to cancel the execution of the method</param>
+  /// <returns>
+  ///   Task representing the asynchronous execution of the method
+  /// </returns>
+  Task SetTaskOwnership(string                                  sessionId,
+                        IList<(string resultId, string taskId)> requests,
```
lemaitre-aneo, line 232: I think ICollection is enough.

Common/src/Storage/Result.cs
```diff
@@ -28,24 +28,22 @@ namespace ArmoniK.Core.Common.Storage;
...
 /// <param name="SessionId">Id of the session that produces and consumes this data</param>
+/// <param name="ResultId">Unique Id of the result</param>
 /// <param name="Name">Name to reference and access this result</param>
 /// <param name="OwnerTaskId">Id of the task that is responsible of generating this result.</param>
 /// <param name="Status">Status of the result (can be Created, Completed or Aborted)</param>
 /// <param name="DependentTasks">List of tasks that depend on this result.</param>
 /// <param name="CreationDate">Date of creation of the current object.</param>
 /// <param name="Data">Data for the current <paramref name="Name" /></param>
 public record Result(string       SessionId,
+                     string       ResultId,
```
lemaitre-aneo, line 39: Why is the resultId not the first field?

Common/src/gRPC/Services/Agent.cs
```diff
@@ -563,6 +565,41 @@ await completionTask.WaitAsync(cancellationToken)
...
+    var results = request.Results.Select(rc => new Storage.Result(request.SessionId,
+                                                                  Guid.NewGuid()
+                                                                      .ToString(),
+                                                                  rc.Name,
+                                                                  "",
+                                                                  ResultStatus.Created,
+                                                                  new List<string>(),
+                                                                  DateTime.UtcNow,
+                                                                  Array.Empty<byte>()))
+                         .ToList();
```
lemaitre-aneo, line 581: Do you need the ToList here? I think resultTable.Create takes an IEnumerable, so you could give the Select as is.

Common/src/gRPC/Services/Submitter.cs
```diff
@@ -672,6 +672,12 @@ await payloadUploadTasks.WhenAll()
                                  cancellationToken)
                     .ConfigureAwait(false);
 
+    await resultTable_.SetTaskOwnership(sessionId,
```
lemaitre-aneo, line 675: I don't understand why it was not there before, and what changed to mandate its use now.

Common/src/gRPC/Services/Submitter.cs
```diff
@@ -703,32 +709,12 @@ private IObjectStorage PayloadStorage(string session)
...
-                                                                                              ResultStatus.Created,
-                                                                                              new List<string>(),
-                                                                                              DateTime.UtcNow,
-                                                                                              Array.Empty<byte>()));
-
-                                           return (Result: resultModel, Req: new IResultTable.ChangeResultOwnershipRequest(intersect,
-                                                                                                                           request.Id));
-                                         });
-
+    var taskDataModels = requests.Select(request => new IResultTable.ChangeResultOwnershipRequest(parentExpectedOutputKeys.Intersect(request.ExpectedOutputKeys),
```
lemaitre-aneo, line 712: I know that it was already there, but this is accidentally quadratic because of the intersection of lists. It should most likely be an intersection of HashSet or something like that.

Common/src/gRPC/Services/Submitter.cs
```diff
@@ -703,32 +709,12 @@ private IObjectStorage PayloadStorage(string session)
...
+                                                                                                  request.Id));
     await resultTable_.ChangeResultOwnership(session,
                                              parentTaskId,
-                                             taskDataModels.Select(tuple => tuple.Req),
+                                             taskDataModels,
                                              cancellationToken)
                       .ConfigureAwait(false);
-
-    await resultTable_.Create(taskDataModels.SelectMany(task => task.Result),
```
lemaitre-aneo, line 730: What was the use of this guy?

Common/tests/Helpers/SimpleAgent.cs
```diff
@@ -54,6 +54,10 @@ public Task FinalizeTaskCreation(CancellationToken cancellationToken)
                                       CancellationToken          cancellationToken)
     => Task.FromResult(new ResultReply());
 
+  public Task<CreateResultsMetaDataResponse> CreateResultsMetaData(CreateResultsMetaDataRequest request,
+                                                                   CancellationToken            cancellationToken)
+    => throw new NotImplementedException();
```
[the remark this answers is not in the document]
lemaitre-aneo: This is a test only agent.

Common/tests/TestBase/ResultTableTestBase.cs
```diff
@@ -750,4 +760,39 @@ public void AddDependentNotExistingResultShouldThrow()
                                                                                 .ConfigureAwait(false));
     }
   }
+
+  [Test]
+  public async Task SetTaskOwnershipShouldSucceed()
```
lemaitre-aneo, line 765: Would it make sense to do this test on both a result with and without an ownerTaskId?

Common/src/Storage/Result.cs
aneojgurhem: SessionId is also the first field in TaskData. I did the same here

Common/src/gRPC/Services/Submitter.cs
aneojgurhem: Beforehand, we were creting the result during task creation so we were creating the result with the field already filled.

Common/src/gRPC/Services/Submitter.cs
aneojgurhem: We can change it in a perf oriented pr

Common/src/gRPC/Services/Submitter.cs
aneojgurhem: create the results during task creation

Common/tests/TestBase/ResultTableTestBase.cs
aneojgurhem: If we override the value, it does not make any difference if there is a value or not. I need to check what is the exact query

lemaitre-aneo: The test I propose would make sure this assumption is correct. I agree the current implementation do not care (and must not care), but I can imagine that some implementation might have make an invalid assumption.
I'm fine both ways, though.

Common/tests/Helpers/SimpleAgent.cs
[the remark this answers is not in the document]
aneojgurhem: Done

Common/src/gRPC/Services/Agent.cs
aneojgurhem: Yes! I'm iterating two times over the results here; first during insertion in DB, second during return. Without the `ToList`, the guid would be generated twice and the result created and the results returned would not be the same.

lemaitre-aneo: Ah. Good call.

dbrasseur-aneo (commented):
👍

Common/src/Storage/IResultTable.cs
aneojgurhem: Done

dbrasseur-aneo (approved):

Common/tests/Helpers/TestDatabaseProvider.cs
```diff
@@ -57,8 +57,18 @@ public TestDatabaseProvider(Action<IServiceCollection>? configurator = null)
...
+
     runner_ = MongoRunner.Run(options);
-    client_ = new MongoClient(runner_.ConnectionString);
+    var settings = MongoClientSettings.FromConnectionString(runner_.ConnectionString);
+    //settings.ClusterConfigurator = cb =>
+    //                               {
+    //                                 cb.Subscribe<CommandStartedEvent>(e => loggerDB.LogTrace("{CommandName} - {Command}",
+    //                                                                                          e.CommandName,
+    //                                                                                          e.Command.ToJson()));
+    //                               };
```
lemaitre-aneo, line 70: If you don't need it, remove it.

lemaitre-aneo (dismissed):

Common/tests/Helpers/TestDatabaseProvider.cs
aneojgurhem: I don't want to see the requests most of the time but I needed it today.

Adaptors/MongoDB/src/ResultTable.cs
```diff
@@ -310,21 +296,39 @@ public class ResultTable : IResultTable
...
 
+  /// <inheritdoc />
+  public async Task SetTaskOwnership(string                                        sessionId,
+                                     ICollection<(string resultId, string taskId)> requests,
+                                     CancellationToken                             cancellationToken = default)
+  {
+    using var activity         = activitySource_.StartActivity($"{nameof(SetTaskOwnership)}");
+    var       resultCollection = resultCollectionProvider_.Get();
+
+    var res = await resultCollection.BulkWriteAsync(requests.Select(r => new UpdateManyModel<Result>(Builders<Result>.Filter.Eq(model => model.ResultId,
```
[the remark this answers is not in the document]
aneojgurhem: I changed it to One

Common/tests/Helpers/TestDatabaseProvider.cs
lemaitre-aneo: Maybe we could have this option more globally. Anyway, approving.

lemaitre-aneo (approved):

PR #313, opened 2023-04-07
refactor: remove IObjectStorageFactory

aneojgurhem:
Remove IObjectStorageFactory since it makes a disctinction between results and payloads which is unnecessary. Moreover, we will soon introduce APIs that will need that there is no distinctions between types of data.


aneojgurhem:
Tasks in a session should only access data in the same session

Adaptors/LocalStorage/src/ObjectStorage.cs
```diff
@@ -47,16 +50,55 @@ public class ObjectStorage : IObjectStorage
...
   }
 
+  /// <inheritdoc />
+  public Task Init(CancellationToken cancellationToken)
+  {
+    _ = cancellationToken;
+    logger_.LogDebug("Initializing Local ObjectStorageFactory at path {path}, chunked by {chunkSize}",
+                     path_,
+                     chunkSize_);
+    Directory.CreateDirectory(path_);
```
lemaitre-aneo, line 76: Does this fail when the directory already exist?

Adaptors/S3/src/ObjectStorage.cs
```diff
@@ -48,16 +52,49 @@ public class ObjectStorage : IObjectStorage
...
+  /// <inheritdoc />
+  public async Task Init(CancellationToken cancellationToken)
+  {
+    if (!isInitialized_)
+    {
+      await AmazonS3Util.DoesS3BucketExistV2Async(s3Client_,
+                                                  bucketName_);
+    }
+
+    logger_.LogInformation("ObjectStorageFactory has correctly been initialized.");
```
lemaitre-aneo, line 73: In the if ?

aneojgurhem: I did not change the implementation of the Init. I took the one from the IObjectStorageFactory

Adaptors/LocalStorage/src/ObjectStorage.cs
lemaitre-aneo:
```suggestion
    // This creates all intermediate directories and does not fail if it already exists: https://learn.microsoft.com/en-us/dotnet/api/system.io.directory.createdirectory
    Directory.CreateDirectory(path_);
```

lemaitre-aneo (approved):

PR #316, opened 2023-04-11
feat: Dependency checking without external service

### LINKED ISSUE #318: Remove dependency checking as external service

aneojgurhem:
Rework data dependency resolution to remove the Dependency checker so that no infrastructure changes are needed.

It implies that there will be more work at the end of the task so the overhead per task will increase.

- [x] remove dependency checker
- [x] abort results in error

fix #318


Adaptors/Memory/src/TaskTable.cs
```diff
@@ -417,6 +417,35 @@ public Task<IEnumerable<PartitionTaskStatusCount>> CountPartitionTasksAsync(Canc
...
 
+  public Task RemoveRemainingDataDependenciesAsync(IEnumerable<(string taskId, IEnumerable<string> dependenciesToRemove)> dependencies,
+                                                   CancellationToken                                                      cancellationToken = default)
+  {
+    using var _ = Logger.LogFunction();
+
+    foreach (var (taskId, dependenciesToRemove) in dependencies)
+    {
+      taskId2TaskData_.AddOrUpdate(taskId,
+                                   _ => throw new InvalidOperationException("The task does not exist."),
```
lemaitre-aneo, line 428: Don't we have a dedicated exception for this error?

Common/tests/Pollster/AgentTest.cs
```diff
@@ -493,13 +485,15 @@ await holder.Agent.FinalizeTaskCreation(CancellationToken.None)
...
-                    holder.QueueStorage.Messages[DependencyResolverOptions.UnresolvedDependenciesQueue]);
+    /*
+     Assert.Contains(holder.TaskWithDependencies1,
+                       holder.QueueStorage.Messages[DependencyResolverOptions.UnresolvedDependenciesQueue]);
     Assert.Contains(holder.TaskWithDependencies2,
                     holder.QueueStorage.Messages[DependencyResolverOptions.UnresolvedDependenciesQueue]);
     Assert.AreEqual(2,
                     holder.QueueStorage.Messages[DependencyResolverOptions.UnresolvedDependenciesQueue]
                           .Count);
+    */
```
dbrasseur-aneo, line 496: To remove

Adaptors/MongoDB/src/TaskTable.cs
```diff
@@ -526,6 +526,26 @@ await foreach (var taskId in taskCollection.AsQueryable(sessionHandle)
     }
   }
 
+  /// <inheritdoc />
+  public async Task RemoveRemainingDataDependenciesAsync(IEnumerable<(string taskId, IEnumerable<string> dependenciesToRemove)> dependencies,
```
lemaitre-aneo, line 530: I am not sure we really need that much flexibility. From what I understand, we need only to remove the exact same results from all tasks.

```suggestion
  public async Task RemoveRemainingDataDependenciesAsync(ICollection<string> taskId, ICollection<string> dependenciesToRemove,
```

I think this would make the request lighter.

Common/src/gRPC/Services/Agent.cs
```diff
@@ -113,20 +114,82 @@ public async Task FinalizeTaskCreation(CancellationToken cancellationToken)
...
+                     .ToList();
+      await pushQueueStorage_.PushMessagesAsync(ids,
+                                                group.Key.PartitionId,
+                                                group.Key.Priority,
+                                                cancellationToken)
+                             .ConfigureAwait(false);
+
+      await taskTable_.FinalizeTaskCreation(ids,
+                                            cancellationToken)
+                      .ConfigureAwait(false);
```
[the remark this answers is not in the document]
aneojgurhem: The conditions in the where (line 176) should reduce this issue. We need to check how often this happens.
Moreover, we already support message deduplication.
Thus, we need to quantify how often it happens. If it does not happen too often overengineering this does not make sense.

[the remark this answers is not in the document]
aneojgurhem: well, message deduplication does not work as well as epected

Adaptors/MongoDB/src/TaskTable.cs
aneojgurhem: As collections are not compatible with lists and arrays, its purpose is unclear to me

lemaitre-aneo: My comment was not really about collection vs enumerable, but about the IEnumerable of IEnumerable. The interface you wrote is more flexible and enable to remove different sets of results depending on the task, but from what I understand, we always remove the same results from all the tasks. Thus, you can have 2 enumerables, one for results to remove, and the other for the tasks. The goal of this change is to make the db request smaller on the network, and less boilerplate to call this function.

Adaptors/Memory/src/TaskTable.cs
lemaitre-aneo: You have change it in another place =P

Adaptors/MongoDB/src/TaskTable.cs
aneojgurhem: I disagree, it depends on the workflow of our users. Today we have mostly one consumer for our data. But it will change in the future, for instance in some algebra based applications. Thus, tasks may complete more than one result at a time and results can be used by more than one task. What you propose means that we will have to remove all the data dependencies from all the tasks even if we know that most of them are not in the list from which we will try to remove them.

lemaitre-aneo:
I have tested this with HTCMock. When the number of dependencies per task is reasonable (1000 tasks 1 level, or 10 000 tasks and at least 10 levels), it is roughly as fast as with the old dependency system (full polling). With a high number of dependencies (10 000 tasks, 1 level), it is roughly 3 times slower on tiny tasks.

I don't think that's a problem for now, but we should keep this in mind and improve it in a future PR.

lemaitre-aneo (approved):

PR #325, opened 2023-04-19
perf: Optimized dependency resolution

lemaitre-aneo:
Previous PR was slowing down dependency resolution in case some tasks had *many* dependencies. This PR solves it by replacing the list of dependencies by an object in MongoDB to avoid linear time to remove elements from the dependencies of a task.
This also solves an increase in memory consumption of the agents.

fix https://github.com/aneoconsulting/ArmoniK/issues/1017


aneojgurhem (commented):
LGTM 👍 
Except for the modified headers

aneojgurhem (approved):

PR #330, opened 2023-04-24
fix: mTLS api and tests in core

dbrasseur-aneo:
Api now supports TLS and mTLS connections. We need to test them in the CI here.
fixes: https://github.com/aneoconsulting/ArmoniK/issues/1023


Tests/Stream/Client/StreamWrapperTests.cs
```diff
@@ -74,12 +74,15 @@ public void TearDown()
   {
     partition_ = null;
     client_    = null;
-    channel_?.Dispose();
+    channel_?.ShutdownAsync()
+            .GetAwaiter()
+            .GetResult();
+    channel_ = null;
```
aneojgurhem, line 80:
```suggestion
    channel_?.ShutdownAsync()
            .Wait();
    channel_ = null;
```

terraform/modules/submitter/locals.tf
```diff
@@ -2,6 +2,7 @@ locals {
   env = [
     "Submitter__DefaultPartition=TestPartition0",
     "Zipkin__Uri=${var.zipkin_uri}",
+    "ASPNETCORE_ENVIRONMENT=Development"
```
aneojgurhem, line 5: Il me semblait qu'elle etait deja présente

.github/workflows/build.yml
```diff
@@ -548,7 +548,82 @@ jobs:
...
+         sudo cp terraform/.terraform/ingress_tls/server/ca.crt /usr/local/share/ca-certificates/ca_tls.crt
+         sudo cp terraform/.terraform/ingress_mtls/server/ca.crt /usr/local/share/ca-certificates/ca_mtls.crt
+         sudo update-ca-certificates
+
+      - name: Run connection test
+        run: CA_INSTALLED=${{ matrix.ca }} dotnet test Tests/Common/Client/src/ArmoniK.Core.Common.Tests.Client.csproj --filter ArmoniK.Core.Common.Tests.Client.ConnectionTest --logger "trx;LogFileName=test-results.trx"
+
+      - name: Test Report
+        uses: dorny/test-reporter@v1
+        if: success() || failure()
```
lemaitre-aneo, line 600:
```suggestion
        if: always()
```

Common/tests/Helpers/GrpcSubmitterServiceHelper.cs
```diff
@@ -115,7 +116,6 @@ public void Dispose()
     server_ = null;
     handler_?.Dispose();
     handler_ = null;
-    channel_?.Dispose();
```
lemaitre-aneo, line 118: Why do you use ShutdownAsync on the tests and not here ?

terraform/main.tf
```diff
@@ -134,3 +134,19 @@ module "partition_metrics_exporter" {
   metrics_env_vars   = module.metrics_exporter.metrics_env_vars
   log_driver         = module.fluenbit.log_driver
 }
+
+module "ingress" {
+  source             = "./modules/ingress"
+  for_each           = tomap({ for c in var.ingress.configs : c.container_name => c })
```
lemaitre-aneo, line 140: You should not need the call to `tomap`

terraform/variables.tf
```diff
@@ -202,3 +202,35 @@ variable "database_image" {
...
+        port           = 5202,
+        tls            = true
+      },
+      {
+        container_name = "ingress_mtls",
+        port           = 5203,
+        tls            = true,
+        mtls           = true
+      }
+  ] }
```
lemaitre-aneo, line 235:
```suggestion
  type = object({
    image = optional(string, "nginxinc/nginx-unprivileged"),
    tag   = optional(string, "1.23.3"),
    configs = map(object({
      port = number,
      tls  = optional(bool, false),
      mtls = optional(bool, false),
    }))
  })
  default = {
    configs = {
      ingess = {
        port = 5201,
      },
      ingess_tls = {
        port = 5201,
        tls  = true,
      },
      ingess_mtls = {
        port = 5201,
        tls  = true,
        mtls = true,
      },
    }
  }
```

.github/workflows/build.yml
dbrasseur-aneo: https://docs.github.com/en/actions/learn-github-actions/expressions#always
Should we ?

terraform/modules/ingress/certificates.tf
```diff
@@ -0,0 +1,179 @@
...
+  password        = ""
+  cert_pem        = tls_locally_signed_cert.ingress_client_certificate.0.cert_pem
+  private_key_pem = tls_private_key.ingress_client_private_key.0.private_key_pem
+  ca_pem          = tls_self_signed_cert.client_root_ingress.0.cert_pem
+}
+
+resource "local_sensitive_file" "ingress_ca" {
+  count           = length(tls_self_signed_cert.root_ingress)
+  content         = tls_self_signed_cert.root_ingress.0.cert_pem
+  filename        = "${path.root}/.terraform/${var.container_name}/server/ca.crt"
```
lemaitre-aneo, line 135: Why do you put these in `.terraform`? This folder should only be used by terraform.

terraform/modules/ingress/certificates.tf
```diff
@@ -0,0 +1,179 @@
...
+  ecdsa_curve = "P384"
+  rsa_bits    = "4096"
+}
+
+resource "tls_cert_request" "ingress_cert_request" {
+  count           = length(tls_private_key.ingress_private_key)
+  private_key_pem = tls_private_key.ingress_private_key.0.private_key_pem
+  subject {
+    country     = "France"
+    common_name = "localhost"
```
lemaitre-aneo, line 70: `localhost` or `armonik.local` or `ingress.armonik`? I don't really care either way.

terraform/modules/ingress/certificates.tf
```diff
@@ -0,0 +1,179 @@
...
+  content         = tls_locally_signed_cert.ingress_client_certificate.0.cert_pem
+  filename        = "${path.root}/.terraform/${var.container_name}/client/client.crt"
+  file_permission = "0644"
+}
+
+resource "local_sensitive_file" "ingress_client_key" {
+  count           = length(tls_private_key.ingress_client_private_key)
+  content         = tls_private_key.ingress_client_private_key.0.private_key_pem
+  filename        = "${path.root}/.terraform/${var.container_name}/client/client.key"
+  file_permission = "0644"
```
lemaitre-aneo, line 171: Keys should not be 644 but 600

terraform/modules/ingress/inputs.tf
```diff
@@ -0,0 +1,47 @@
...
+  type = bool
+}
+
+variable "submitter_url" {
+  type = string
+}
+
+variable "submitter_port" {
+  type = number
+}
```
lemaitre-aneo, line 35: You could group those 2:

```suggestion
variable "submitter" {
  type = object({
    url  = string,
    port = number,
  })
}
```

I think you could do the same with some other parameters.

terraform/modules/ingress/locals.tf
```diff
@@ -0,0 +1,24 @@
...
+        source = abspath("${path.root}/.terraform/${var.container_name}/server/ingress.key")
+    }] : [],
+    var.mtls ? [{
+      target = "/client_ca.crt"
+      source = abspath("${path.root}/.terraform/${var.container_name}/client/client_ca.crt")
+    }] : []
+  )
+  volume_map = tomap({
+    for t in local.volumes_list : t.target => t
+  })
```
lemaitre-aneo, line 23:
```suggestion
  volumes = merge({
      "/etc/nginx/conf.d/default.conf" = abspath(local_file.ingress_conf_file.filename),
    }, !var.tls ? {} : {
      "/ingress.crt" = abspath("${path.root}/.terraform/${var.container_name}/server/ingress.crt"),
      "/ingress.key" = abspath("${path.root}/.terraform/${var.container_name}/server/ingress.key"),
    }, !var.mtls ? {} : {
      "/client_ca.crt" = abspath("${path.root}/.terraform/${var.container_name}/client/client_ca.crt"),
    }
  }
```

terraform/modules/ingress/certificates.tf
dbrasseur-aneo: For SSL validation it's important, so unless I modify the hosts file, localhost is mandatory

terraform/modules/ingress/certificates.tf
dbrasseur-aneo: Well, I've had a lot of issues with that. Because we're using nginx unpriviledged, mounting the file directly in docker keeps its permissions, and so the server can't access it if it's not 644

lemaitre-aneo (dismissed):

.github/workflows/build.yml
lemaitre-aneo: I think, yes. But I don't mind keeping it this way.

lemaitre-aneo (approved):

PR #332, opened 2023-04-25
refactor: attempt to catch the InnerException to have more logs

melflitty-aneo:


aneojgurhem:
@melflitty I think we can close it now. Are you ok ?

PR #333, opened 2023-04-26
feat: implement rpc methods to create tasks and results independently

aneojgurhem:
- [x] refactor task creation to extract common code that will be used in both submission ways (task and payload together vs task and payload separately)
- [x] refactor task options conversion between storage and gRPC API version to properly manage nullables
- [x] implement new rpc method to create tasks and payloads separately
- [ ] use the new methods in our bench to validate their implementation (todo in another pr)
- [x] tests new methods in IAgent interface
- [x] tests new methods in IResultTable interface

### DIFF
--- a/Common/src/Storage/TaskCreationRequest.cs
+++ b/Common/src/Storage/TaskCreationRequest.cs
@@ -1,4 +1,4 @@
-﻿// This file is part of the ArmoniK project
+// This file is part of the ArmoniK project
 // 
 // Copyright (C) ANEO, 2021-2023. All rights reserved.
 // 
@@ -19,6 +19,8 @@ using System.Collections.Generic;
 
 namespace ArmoniK.Core.Common.Storage;
 
-public record TaskRequest(string              Id,
-                          IEnumerable<string> ExpectedOutputKeys,
-                          IEnumerable<string> DataDependencies);
+public record TaskCreationRequest(string              TaskId,
+                                  string              PayloadId,
+                                  TaskOptions         Options,
+                                  ICollection<string> ExpectedOutputKeys,
+                                  ICollection<string> DataDependencies);
--- a/Common/src/Storage/TaskLifeCycleHelper.cs
+++ b/Common/src/Storage/TaskLifeCycleHelper.cs
@@ -1,0 +1,237 @@
+// This file is part of the ArmoniK project
+// 
+// Copyright (C) ANEO, 2021-2023. All rights reserved.
+// 
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as published
+// by the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+// 
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY, without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+// 
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+
+using ArmoniK.Api.Common.Utils;
+using ArmoniK.Api.gRPC.V1;
+using ArmoniK.Core.Base;
+using ArmoniK.Core.Common.Exceptions;
+using ArmoniK.Utils;
+
+using Grpc.Core;
+
+using Microsoft.Extensions.Logging;
+
+using TaskStatus = ArmoniK.Api.gRPC.V1.TaskStatus;
+
+namespace ArmoniK.Core.Common.Storage;
+
+public static class TaskLifeCycleHelper
+{
+  public static TaskOptions ValidateSession(SessionData       sessionData,
+                                            TaskOptions?      submissionOptions,
+                                            string            parentTaskId,
+                                            int               maxPriority,
+                                            ILogger           logger,
+                                            CancellationToken cancellationToken)
+  {
+    var localOptions = submissionOptions != null
+                         ? TaskOptions.Merge(submissionOptions,
+                                             sessionData.Options)
+                         : sessionData.Options;
+
+    using var logFunction = logger.LogFunction(parentTaskId);
+    using var sessionScope = logger.BeginPropertyScope(("Session", sessionData.SessionId),
+                                                       ("TaskId", parentTaskId),
+                                                       ("PartitionId", localOptions.PartitionId));
+
+    if (logger.IsEnabled(LogLevel.Trace))
+    {
+      cancellationToken.Register(() => logger.LogTrace("CancellationToken from ServerCallContext has been triggered"));
+    }
+
+    var availablePartitionIds = sessionData.PartitionIds ?? Array.Empty<string>();
+    if (!availablePartitionIds.Contains(localOptions.PartitionId))
+    {
+      throw new InvalidOperationException($"The session {sessionData.SessionId} is assigned to the partitions " +
+                                          $"[{string.Join(", ", availablePartitionIds)}], but TaskRequest is assigned to partition {localOptions.PartitionId}");
+    }
+
+    if (localOptions.Priority >= maxPriority)
+    {
+      var exception = new RpcException(new Status(StatusCode.InvalidArgument,
+                                                  $"Max priority is {maxPriority}"));
+      logger.LogError(exception,
+                      "Invalid Argument");
+      throw exception;
+    }
+
+    return localOptions;
+  }
+
+  public static async Task CreateTasks(ITaskTable                       taskTable,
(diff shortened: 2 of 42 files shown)


Common/src/gRPC/Services/GrpcResultsService.cs
```diff
@@ -125,4 +131,138 @@ public class GrpcResultsService : Results.ResultsBase
...
+                                                                                       }.ToAsyncEnumerable(),
+                                                                                       context.CancellationToken)
+                                                                     .ConfigureAwait(false);
+
+                                                 return result;
+                                               })
+                               .WhenAll()
+                               .ConfigureAwait(false);
+
+    var resultList = await results.Select(async r => await resultTable_.CompleteResult(request.SessionId,
```
lemaitre-aneo, line 165: I think you could "merge" this select with the previous one. Besides, you could use ParallelSelect if you want.

Common/src/gRPC/Services/Agent.cs
```diff
@@ -644,6 +642,152 @@ await completionTask.WaitAsync(cancellationToken)
            };
   }
 
+  /// <inheritdoc />
+  public async Task<SubmitTasksResponse> SubmitTasks(SubmitTasksRequest request,
```
lemaitre-aneo, line 646: I am not sure why do you need to duplicate those methods from submitter.

Common/src/Storage/TaskCreationRequest.cs
```diff
@@ -19,6 +19,8 @@
 
 namespace ArmoniK.Core.Common.Storage;
 
-public record TaskRequest(string              Id,
-                          IEnumerable<string> ExpectedOutputKeys,
-                          IEnumerable<string> DataDependencies);
+public record TaskCreationRequest(string              TaskId,
+                                  string              PayloadId,
+                                  TaskOptions         Options,
```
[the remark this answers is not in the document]
aneojgurhem: Not here, by design, we intend to put task options in the task table that come from the session if null during submission. As we check the options before creating this request, it does not make sense to make it nullable. On the opposite, It will make it less clear.
Refactor options management is not a goal of this development and can be done in another PR.

Common/src/Storage/TaskOptions.cs
```diff
@@ -100,3 +75,42 @@ public record TaskOptions(IDictionary<string, string> Options,
...
+           taskOption.ApplicationService,
+           taskOption.EngineType);
+
+  public static TaskOptions? ToNullableTaskOptions(this Api.gRPC.V1.TaskOptions? taskOption)
+    => taskOption?.ToTaskOptions();
+}
+
+public static class TaskOptionsExt
+{
+  public static Api.gRPC.V1.TaskOptions ToTaskOptions(this TaskOptions taskOption)
```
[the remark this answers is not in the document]
aneojgurhem: and break nullable. It would generate a lot of warnings.

And there is a great issue I encountered:
- CreateTask takes an option and an `IAsyncEnumerable<TaskRequest>`
- the conversion from gprc options to internal options does not work because the grpc options is null. It comes from the fact that the implicit conversion does not work well with null (we cannot easily detect that the object is null before the implicit conversion and avoid making this conversion that does not work. Doing so takes a lot of lines of code that are not really self-explanatory)
- so implicit conversion does not work but we still enters task creation through the IAsyncEnumerable
- request are not consummed and application wait
- task creation handle indefinitely.

Common/src/Storage/TaskOptions.cs
```diff
@@ -33,39 +33,14 @@ public record TaskOptions(IDictionary<string, string> Options,
                           string                      ApplicationService,
                           string                      EngineType)
 {
-  public static implicit operator Api.gRPC.V1.TaskOptions(TaskOptions taskOption)
```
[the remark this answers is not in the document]
aneojgurhem: cf issue mentionned above

Common/src/gRPC/Services/Agent.cs
aneojgurhem: because this one does not have to manage payload data

Common/src/gRPC/Services/GrpcResultsService.cs
aneojgurhem: parallel select was not existing when I wrote the code.

aneojgurhem: well there is an issue with this code. Results are never created in db

Common/src/Storage/TaskOptions.cs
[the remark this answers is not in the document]
lemaitre-aneo: Here is the comment:
> and break nullable. It would generate a lot of warnings.
> 
> And there is a great issue I encountered:
> 
>     * CreateTask takes an option and an `IAsyncEnumerable<TaskRequest>`
> 
>     * the conversion from gprc options to internal options does not work because the grpc options is null. It comes from the fact that the implicit conversion does not work well with null (we cannot easily detect that the object is null before the implicit conversion and avoid making this conversion that does not work. Doing so takes a lot of lines of code that are not really self-explanatory)
> 
>     * so implicit conversion does not work but we still enters task creation through the IAsyncEnumerable
> 
>     * request are not consummed and application wait
> 
>     * task creation handle indefinitely.

Fair enough, I just wanted to know if there was any technical reason or just style preference.

Common/src/Storage/TaskOptions.cs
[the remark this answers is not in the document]
lemaitre-aneo: I think you missed my point here. I just suggested to rename the method from `ToTaskOptions` into `ToGrpcTaskOptions`. I said nothing about nullable. The question mark to ask a question, not making the call nullable.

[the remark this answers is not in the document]
aneojgurhem: oh right ! I will change the name to make it clearer !

Common/src/Storage/TaskCreationRequest.cs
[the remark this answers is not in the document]
lemaitre-aneo: Good call. I missed that the merge with the session TaskOption was already done at this point in time.

Common/tests/Submitter/ExceptionInterceptorTests.cs
```diff
@@ -263,7 +277,7 @@ await foreach (var req in requests.WithCancellation(cancellationToken)
                                                                      SessionId = "SessionId",
                                                                      TaskOptions = new TaskOptions
                                                                                    {
-                                                                                     MaxDuration = Duration.FromTimeSpan(TimeSpan.MaxValue),
+                                                                                     MaxDuration = Duration.FromTimeSpan(TimeSpan.FromSeconds(1)),
```
lemaitre-aneo, line 280: If you want an Infinite timespan, you could use `Timeout.Infinite`

aneojgurhem: I just need a valid timespan for grpc. 1 second is okay.

Common/src/Storage/TaskLifeCycleHelper.cs
```diff
@@ -0,0 +1,237 @@
...
+
+    var taskDataModels = taskRequests.Select(request => new IResultTable.ChangeResultOwnershipRequest(parentExpectedOutputKeys.Intersect(request.ExpectedOutputKeys),
+                                                                                                      request.TaskId));
+    await resultTable.ChangeResultOwnership(sessionId,
+                                            parentTaskId,
+                                            taskDataModels,
+                                            cancellationToken)
+                     .ConfigureAwait(false);
+
+    var readyTasks = new List<(string taskId, int priority, string partitionId)>();
```
lemaitre-aneo, line 156: Maybe `readyTasks` could be a `Dictionary<(string parittionId, int priority), string>`. This would provide the group by naturally during construction

aneojgurhem: but (partitionid, priority) is not unique

Common/src/Storage/TaskLifeCycleHelper.cs
```diff
@@ -0,0 +1,237 @@
...
+                         : sessionData.Options;
+
+    using var logFunction = logger.LogFunction(parentTaskId);
+    using var sessionScope = logger.BeginPropertyScope(("Session", sessionData.SessionId),
+                                                       ("TaskId", parentTaskId),
+                                                       ("PartitionId", localOptions.PartitionId));
+
+    if (logger.IsEnabled(LogLevel.Trace))
+    {
+      cancellationToken.Register(() => logger.LogTrace("CancellationToken from ServerCallContext has been triggered"));
```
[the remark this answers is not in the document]
aneojgurhem: I agree.
I took the existing code. I just moved it.
Let's create an issue

Common/src/Storage/TaskLifeCycleHelper.cs
lemaitre-aneo: Sorry, I meant `List<string>`. Go to another PR.

lemaitre-aneo (approved):

PR #334, opened 2023-04-26
refactor: Change transient to singleton

ddiakiteaneo:
Change transient to a singleton in MongoDB client.


Adaptors/MongoDB/src/Common/SessionProvider.cs
```diff
@@ -40,6 +40,7 @@ public class SessionProvider : IInitializable
   public SessionProvider(IMongoClient client)
     => client_ = client;
 
+
```
aneojgurhem, line 43: This line should not be modified

aneojgurhem (changes requested):

aneojgurhem (approved):

PR #339, opened 2023-05-04
perf: improve mongodb requests using native mongodb find instead of linq queries

alouvion-aneo:
[retlated CR](https://github.com/aneoconsulting/ArmoniK/issues/1057)
idea here is to use dottrace to check if we can improve mongo db requests

basically we can replace :
```
AsQueryable(sessionHandle).Where(sdm => sdm.SessionId == sessionId)
```
by :
`.Find(session => session.SessionId == sessionId)
`

You can notice 2x to 3x perfs in the mongoDB calls :
image

we are always performing the same requests with different parameters so :
=> I think further improvements may be to use directly mongo syntax with prebuilt requests
=> seems ugly ok but it could be good to check perf diff with this


aneojgurhem:
That's a perf: contribution !

aneojgurhem:
I think that there are more instances of this issue than that. Could you check ?

aneojgurhem (approved):

PR #341, opened 2023-05-05
chore: add some useful scripts

aneojgurhem:
Add scripts for
- update api from command line (python)
- restart failed pipelines jobs (gh cli + bash)

### DIFF
--- a/tools/update.py
+++ b/tools/update.py
@@ -1,0 +1,30 @@
+from re import compile
+import subprocess
+import argparse
+
+parser = argparse.ArgumentParser(
+    description="Set the version of the ArmoniK.Api* packages to the new version",
+    formatter_class=argparse.ArgumentDefaultsHelpFormatter,
+)
+parser.add_argument("new_version", help="New version", type=str)
+args = parser.parse_args()
+
+cmd = subprocess.Popen('git ls-tree -r HEAD --name-only --full-tree | grep "csproj$"', shell=True, stdout=subprocess.PIPE, text=True)
+cmd.wait()
+
+files = cmd.communicate()[0].rstrip("\n").split("\n")
+
+regex = compile(r'(<\s*PackageReference\s+Include\s*=\s*"ArmoniK\.Api\.[a-zA-Z]+"\s+Version\s*=\s*").*("\s*\/>)')
+
+for file in files:
+    print(file)
+
+    lines = []
+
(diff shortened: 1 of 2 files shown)


lemaitre-aneo (commented):
I would add some comments to explain what those scripts do. Ideally behind a `-h` flag.

aneojgurhem:
> I would add some comments to explain what those scripts do. Ideally behind a `-h` flag.

That's already the case for the python script.
Do you have some suggestions for the bash script ?

tools/update.py
```diff
@@ -0,0 +1,30 @@
...
+
+    lines = []
+
+    with open(file, "r") as f:
+        for line in f.readlines():
+            newline = regex.sub(fr'\g<1>{args.new_version}\g<2>', line)
+            lines.append(newline)
+
+    with open(file, "w") as f:
+        f.writelines(lines)
```
dbrasseur-aneo, line 30:
```suggestion
    with open(file, "r+") as f:
        lines = [regex.sub(fr'\g<1>{args.new_version}\g<2>', line) for line in f.readlines()]
        f.seek(0)
        f.writelines(lines)
```

dbrasseur-aneo (commented):
Maybe with some comments as well to iunderstand the script better

tools/update.py
lemaitre-aneo: I'm pretty sure you also need to truncate the file in case the new content is shorter than the previous.

tools/update.py
```diff
@@ -0,0 +1,25 @@
...
+files = cmd.communicate()[0].rstrip("\n").split("\n")
+
+regex = compile(r'(<\s*PackageReference\s+Include\s*=\s*"ArmoniK\.Api\.[a-zA-Z]+"\s+Version\s*=\s*").*("\s*\/>)')
+
+for file in files:
+    print(file)
+
+    with open(file, "r+") as f:
+        lines = [regex.sub(fr'\g<1>{args.new_version}\g<2>', line) for line in f.readlines()]
+        f.seek(0)
```
dbrasseur-aneo, line 24:
```suggestion
        f.truncate(0)
        f.seek(0)
```

tools/update.py
aneojgurhem: yes that's exactly my issue

dbrasseur-aneo (approved):

PR #342, opened 2023-05-05
feat: Add task duration in database

aneojgurhem:
- [x] Add task duration in database
- [x] Use new version of ArmoniK.Api that expects the task duration in several RPCs

### DIFF
--- a/Adaptors/Memory/src/TaskTable.cs
+++ b/Adaptors/Memory/src/TaskTable.cs
@@ -131,22 +131,22 @@ public class TaskTable : ITaskTable
   }
 
   /// <inheritdoc />
-  public Task StartTask(string            taskId,
+  public Task StartTask(TaskData          taskData,
                         CancellationToken cancellationToken = default)
   {
-    if (!taskId2TaskData_.ContainsKey(taskId))
+    if (!taskId2TaskData_.ContainsKey(taskData.TaskId))
     {
-      throw new TaskNotFoundException($"Key '{taskId}' not found");
+      throw new TaskNotFoundException($"Key '{taskData.TaskId}' not found");
     }
 
-    taskId2TaskData_.AddOrUpdate(taskId,
+    taskId2TaskData_.AddOrUpdate(taskData.TaskId,
                                  _ => throw new InvalidOperationException("The task does not exist."),
                                  (_,
                                   data) => data with
                                            {
                                              Status = TaskStatus.Processing,
-                                             StartDate = DateTime.UtcNow,
-                                             PodTtl = DateTime.UtcNow,
+                                             StartDate = taskData.StartDate,
+                                             PodTtl = taskData.PodTtl,
                                            });
     return Task.CompletedTask;
   }
@@ -380,7 +380,7 @@ public class TaskTable : ITaskTable
   }
 
   /// <inheritdoc />
-  public Task SetTaskSuccessAsync(string            taskId,
+  public Task SetTaskSuccessAsync(TaskData          taskData,
                                   CancellationToken cancellationToken = default)
   {
     using var _ = Logger.LogFunction();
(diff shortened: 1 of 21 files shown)


Adaptors/Memory/src/TaskTable.cs
```diff
@@ -539,12 +539,9 @@ public Task<IEnumerable<PartitionTaskStatusCount>> CountPartitionTasksAsync(Canc
   }
 
   /// <inheritdoc />
-  public Task<TaskData> AcquireTask(string            taskId,
-                                    string            ownerPodId,
```
lemaitre-aneo, line 543: I'm not a fan of passing the _new_ `ownerPodId` implicitely in `taskData`.

Adaptors/Memory/src/TaskTable.cs
```diff
@@ -131,22 +131,22 @@ public class TaskTable : ITaskTable
...
+    taskId2TaskData_.AddOrUpdate(taskData.TaskId,
                                  _ => throw new InvalidOperationException("The task does not exist."),
                                  (_,
                                   data) => data with
                                            {
                                              Status = TaskStatus.Processing,
-                                             StartDate = DateTime.UtcNow,
-                                             PodTtl = DateTime.UtcNow,
+                                             StartDate = taskData.StartDate,
+                                             PodTtl = taskData.PodTtl,
```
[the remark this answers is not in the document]
aneojgurhem: and throw ?

Common/src/Storage/ITaskTable.cs
```diff
@@ -266,13 +270,20 @@ public interface ITaskTable : IInitializable
...
   ///   Change the status of the task to succeeded
-  /// </summary>
-  /// <param name="taskId">Id of the task to tag as succeeded</param>
+  ///   Updates:
+  ///   - <see cref="TaskData.Status" />
+  ///   - <see cref="TaskData.StartDate" />
+  ///   - <see cref="TaskData.EndDate" />
+  ///   - <see cref="TaskData.CreationToEndDuration" />
+  ///   - <see cref="TaskData.ProcessingToEndDuration" />
+  ///   - <see cref="TaskData.Output" />
```
lemaitre-aneo, line 279: This should most likely be in `<details>` instead of `<summary>`

Common/src/Storage/ITaskTable.cs
```diff
@@ -326,32 +351,36 @@ public interface ITaskTable : IInitializable
 
   /// <summary>
   ///   Acquire the task to process it on the current agent
-  /// </summary>
-  /// <param name="taskId">Id of the task to acquire</param>
-  /// <param name="ownerPodId">Identifier (Ip) that will be used to reach the pod if another pod tries to acquire the task</param>
```
lemaitre-aneo, line 331: This information must be kept in another form.

Common/src/Storage/ITaskTable.cs
aneojgurhem: Details are not shown in the tooltip. I don't like this idea because I need this information

lemaitre-aneo: Sorry, I mixed with `remarks` which is what Microsoft use. I checked and remarks work on Visual Studio.

Adaptors/Memory/src/TaskTable.cs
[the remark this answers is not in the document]
lemaitre-aneo: After more thoughts, I think you could keep the code like that. Especially since it is documented.

Adaptors/Memory/src/TaskTable.cs
lemaitre-aneo: Fair enough. Now that it is documented, it is not as problematic.

lemaitre-aneo (approved):

PR #343, opened 2023-05-09
feat: use .Net 7

aneojgurhem:


melflitty:
Don't we need to change the version of dotnet in the pipeline?

aneojgurhem:
Good point !

PR #346, opened 2023-05-11
refactor: print version of core in logs

aneojgurhem:
Enrich logs with Core's version.

### DIFF
--- a/Common/tests/LoggerExtTest.cs
+++ b/Common/tests/LoggerExtTest.cs
@@ -1,0 +1,56 @@
+// This file is part of the ArmoniK project
+// 
+// Copyright (C) ANEO, 2021-2023. All rights reserved.
+// 
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as published
+// by the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+// 
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY, without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+// 
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+using ArmoniK.Core.Common.gRPC;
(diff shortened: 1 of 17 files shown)


Common/tests/LoggerExtTest.cs
```diff
@@ -0,0 +1,56 @@
...
+    var logger         = loggerProvider.CreateLogger("root");
+
+    logger.LogVersion(typeof(LoggerExtTest));
+    logger.LogVersion(typeof(gRPC.Services.Submitter));
+
+    Assert.Pass();
+  }
+
+  [Test]
+  public void LoggerInithouldSucceed()
```
ngruelaneo, line 45: I think that the name is not the one wanted

Utils/src/LoggerExt.cs
```diff
@@ -0,0 +1,41 @@
...
+  ///   Logs information about the given type
+  /// </summary>
+  /// <param name="logger">Logger used to print the information</param>
+  /// <param name="type">The type to print information about</param>
+  /// <param name="logLevel">The level used to print the information</param>
+  public static void LogVersion(this ILogger logger,
+                                Type         type,
+                                LogLevel     logLevel = LogLevel.Information)
+    => logger.Log(logLevel,
+                  "Version of {class} is {version} ({qualifiedName})",
```
ngruelaneo, line 36: I will suggest another format: 

`logger.Log(logLevel,
            class,
            version,
...
`

I do not know what is qualifiedName but it can be added also.
The reason it that this log format will be much easier to parse than a string

aneojgurhem: raw logs look like this:

```json
{"@t":"2023-05-12T09:11:59.7750437Z","@mt":"Version of {class} is {version} ({qualifiedName})","class":"ArmoniK.Core.Common.Tests.LoggerExtTest","version":"1.0.0.0","qualifiedName":"ArmoniK.Core.Common.Tests.LoggerExtTest, ArmoniK.Core.Common.Tests, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null","SourceContext":"root","CoreVersion":"1.0.0.0"}

```

aneojgurhem: So "parsing" is loading a JSON dictionnary

ngruelaneo (approved):

PR #347, opened 2023-05-15
refactor: externalize adapter loading test

melflitty-aneo:


lemaitre-aneo:
@melflitty Does it solve your issue to use an external plugin?

PR #348, opened 2023-05-17
fix: give valid version for packages in justfile when not specified

melflitty:


aneojgurhem (commented):
I'd put `0.0.0.0-local` so that it does not conflict with anything in the future
I'd do 2 prs too^^

aneojgurhem (approved):

PR #359, opened 2023-05-26
refactor: implement SetTaskError, SetTaskCancelled and SetTaskCompleted as extensions method of UpdateOneTask

aneojgurhem:


Common/src/Storage/ITaskTable.cs
```diff
@@ -251,6 +251,20 @@ public interface ITaskTable : IInitializable
                                          Expression<Func<TaskData, T>>    selector,
                                          CancellationToken                cancellationToken = default);
 
+  // TODO Should be compatible with EFCORE : https://learn.microsoft.com/en-us/ef/core/saving/execute-insert-update-delete#updating-multiple-properties
```
dbrasseur-aneo, line 254: If this to do in this PR or in another ?

dbrasseur-aneo (approved):

Common/src/Storage/ITaskTable.cs
aneojgurhem: It should be working with EF Core but the adapter for ef core will be in another pr

Common/src/Storage/TaskData.cs
```diff
@@ -151,6 +154,24 @@ public record TaskData(string        SessionId,
   {
   }
 
+  /// <summary>
+  ///   Copy a <see cref="TaskData" /> and modify it according to given updates
```
ngruelaneo, line 158: why do you need to do a copy ?

ngruelaneo (approved):

PR #377, opened 2023-06-05
build: support for external queue

aneojgurhem:
Arbitrary configuration can be passed to terraform through the variable `TF_VAR_custom_env_vars`. The queue can also be set up to none meaning that configurations for an external queue can be passed with the `TF_VAR_custom_env_vars` env var.

example:

```bash
TF_VAR_custom_env_vars='{ "MyQueueEndpoint" = "endpoint", "MyQueuePass" = "pass", "MyQueueUser" = "user" }' just queue=none plan
```


dbrasseur-aneo (approved):
To specify the queue it's with the custom env variables ?

ngruelaneo (approved):

PR #411, opened 2023-06-22
refactor: add more information in the pushed messages including session id

melflitty:


Base/src/DataStructures/MessageData.cs
```diff
@@ -0,0 +1,28 @@
...
+
+namespace ArmoniK.Core.Base.DataStructures;
+
+/// <summary>
+///   Data structure to hold message data
+/// </summary>
+/// <param name="Message">Content of the message</param>
+/// <param name="SessionId">Unique name of the session to which this message belongs</param>
+/// <param name="Options">Task options</param>
+public record MessageData(string      Message,
```
aneojgurhem, line 26: This is the taskId !

Adaptors/Amqp/src/PushQueueStorage.cs
```diff
@@ -108,16 +110,22 @@ public class PushQueueStorage : QueueStorage, IPushQueueStorage
...
 
-      await Task.WhenAll(messages.Select(id => sender.SendAsync(new Message(Encoding.UTF8.GetBytes(id))
-                                                                {
-                                                                  Header = new Header
-                                                                           {
-                                                                             Priority = (byte)internalPriority,
-                                                                           },
-                                                                  Properties = new Properties(),
-                                                                })))
+      await Task.WhenAll(messages.Select(msgData => sender.SendAsync(new Message(Encoding.UTF8.GetBytes(JsonSerializer.Serialize(new
```
aneojgurhem, line 113: As we return the taskId only, I would only send the taskId in.

aneojgurhem (changes requested):

Base/src/DataStructures/MessageData.cs
melflitty: Do i need to rename the variable Message to taskId or just change the doc or both?

lemaitre-aneo: I would say both

Adaptors/Amqp/src/PushQueueStorage.cs
aneojgurhem: We can bypass them based on our needs.

Base/src/DataStructures/MessageData.cs
aneojgurhem: Both

Adaptors/Amqp/src/PushQueueStorage.cs
aneojgurhem: Tests are not working because of that. During the pull, we need to get the task id. You changed the serialization but you did not change the deserialization part !

Base/src/IPushQueueStorage.cs
```diff
@@ -36,8 +38,8 @@ public interface IPushQueueStorage : IQueueStorage
...
   ///   Task representing the asynchronous execution of the method
   /// </returns>
-  public Task PushMessagesAsync(IEnumerable<string> messages,
-                                string              partitionId,
-                                int                 priority          = 1,
-                                CancellationToken   cancellationToken = default);
+  public Task PushMessagesAsync(IEnumerable<MessageData> messages,
+                                string                   partitionId,
+                                int                      priority          = 1,
+                                CancellationToken        cancellationToken = default);
```
lemaitre-aneo, line 44: I wonder. Should the `priority` and `partitionId` be kept as their own parameters? or integrated directly into the `MessageData` structure? I think I would prefer the latter.

@aneojgurhem What do you think?

Base/src/DataStructures/MessageData.cs
```diff
@@ -0,0 +1,28 @@
...
+
+/// <summary>
+///   Data structure to hold message data
+/// </summary>
+/// <param name="TaskId">Unique identifier of the task</param>
+/// <param name="SessionId">Unique name of the session to which this message belongs</param>
+/// <param name="Options">Task options</param>
+public record MessageData(string      TaskId,
+                          string      SessionId,
+                          TaskOptions Options);
```
lemaitre-aneo, line 28: I think it could be nice to have an _explicit_ conversion from a `TaskData`.

Base/src/IPushQueueStorage.cs
aneojgurhem: It will change the meaning of the interface and delegate the management of the priority/partition to the implementation. BTW, these informations are already present in MessageData.TaskOptions

Base/src/DataStructures/MessageData.cs
aneojgurhem: I'm not sure it is really needed

aneojgurhem (dismissed):

Base/src/DataStructures/MessageData.cs
melflitty: What do you mean by that?

aneojgurhem: He means write an extension method `ToMessageData` for TaskData that convert one into the other.

Common/src/gRPC/Services/Agent.cs
```diff
@@ -168,9 +171,13 @@ public async Task FinalizeTaskCreation(CancellationToken cancellationToken)
     {
       var ids = group.Select(data => data.TaskId)
                      .ToList();
-      await pushQueueStorage_.PushMessagesAsync(ids,
+
+      var msgsData = group.Select(data => new MessageData(data.TaskId,
+                                                          data.SessionId,
+                                                          data.Options))
+                          .ToList();
```
aneojgurhem, line 178: and this one

Common/src/Storage/TaskLifeCycleHelper.cs
```diff
@@ -280,11 +289,11 @@ public static class TaskLifeCycleHelper
...
       {
         await pushQueueStorage.PushMessagesAsync(item.Value,
-                                                 item.Key.partitionId,
-                                                 item.Key.priority,
+                                                 item.Key,
                                                  cancellationToken)
                               .ConfigureAwait(false);
-        await taskTable.FinalizeTaskCreation(item.Value,
+        await taskTable.FinalizeTaskCreation(item.Value.Select(data => data.TaskId)
+                                                 .ToList(),
```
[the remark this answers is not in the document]
melflitty: It seems that this one is mandatory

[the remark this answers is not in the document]
aneojgurhem: Why ? Is it an ICollection ?

aneojgurhem (approved):

PR #422, opened 2023-06-28
fix: build and push images in the release pipeline

aneojgurhem:


dbrasseur-aneo (commented):
Will this push the images ? Shouldn't we just add a push step ?

aneojgurhem:
This will build amd64 and arm images then push them

dbrasseur-aneo (approved):

PR #436, opened 2023-07-07
feat: API includes more fields to sort on during sessions list

aneojgurhem:


Common/src/ArmoniK.Core.Common.csproj
```diff
@@ -26,7 +26,7 @@
 
 
   <ItemGroup>
-    <PackageReference Include="ArmoniK.Api.Core" Version="3.10.0" />
+    <PackageReference Include="ArmoniK.Api.Core" Version="3.11.0-feataddtaskoptionstosession.23.58da3e7" />
```
ddiakiteaneo, line 29: Can this API version go on main?

aneojgurhem: Not yet, we need to merge the new Api first

aneojgurhem:
This PR lacks some tests on the usage of the new fields

aneojgurhem:
I added the tests

Common/tests/TestBase/SessionTableTestBase.cs
```diff
@@ -454,4 +443,38 @@ public async Task ListSessionAsyncNoFilterShouldSucceed()
                       res.Count);
     }
   }
+
+
+  [Test]
+  public async Task ListSessionAsyncTaskOptionsOptions()
```
dbrasseur-aneo, line 449: This tests the number of sessions, not that the sort was done correctly

aneojgurhem: Yes, that is the role of the database.

aneojgurhem: It mainly tests that the request is valid

dbrasseur-aneo (approved):

PR #442, opened 2023-07-11
refactor: update Api fields to propagate changes due to oneof fixes

melflitty:


Common/tests/ListApplicationsRequestExt/ToApplicationFieldTest.cs
```diff
@@ -84,13 +84,25 @@ public static IEnumerable<TestCaseData> TestCasesInvoke()
       => new TestCaseData(field,
                           expected).SetArgDisplayNames(field.ToString());
 
-    yield return Case(ApplicationRawField.Service,
+    yield return Case(new ApplicationRawField
```
aneojgurhem, line 87: You should keep one liner `Case` and put the `new` par the in definition of `Case`. Then `Case` would take `ApplicationRawEnumField` as input.
This should be done for every generated testcase based on this idea.

aneojgurhem (changes requested):

aneojgurhem (approved):

PR #443, opened 2023-07-19
feat: implement api filters

aneojgurhem:


Common/src/gRPC/EnumFieldExt.cs
```diff
@@ -0,0 +1,219 @@
...
+  public static Expression<Func<SessionData, object?>> ToField(this SessionRawEnumField enumField)
+    => enumField switch
+       {
+         SessionRawEnumField.SessionId    => session => session.SessionId,
+         SessionRawEnumField.Status       => session => session.Status,
+         SessionRawEnumField.PartitionIds => session => session.PartitionIds,
+         SessionRawEnumField.Options      => session => session.Options,
+         SessionRawEnumField.CreatedAt    => session => session.CreationDate,
+         SessionRawEnumField.CancelledAt  => session => session.CancellationDate,
+         SessionRawEnumField.Duration     => throw new ArgumentOutOfRangeException(),
```
melflitty, line 123: @aneojgurhem Why is SessionRawEnumField.Duration out of range? I mean what is the point of having duration in Api field but not in the corresponding field in Core?

aneojgurhem: It is not implemented yet in the SessionData. We did not implement duration management

Common/src/gRPC/ExpressionBuilders.cs
```diff
@@ -0,0 +1,168 @@
...
+                                                             Type                           type,
+                                                             string                         method,
+                                                             bool                           invert = false)
+  {
+    var member = GetMemberExpression(selector);
+
+    var constant = Expression.Constant(value,
+                                       typeof(T));
+
+    var methodInfo = type.GetMethods(BindingFlags.Static | BindingFlags.Public)
```
melflitty, line 107: Why do you make sure here that the method we get is static and public but we don't do it in the case of string type?

aneojgurhem: Public and static methods are usually extensions methods that will take two parameters : the collection and the value.
For strings, the methods we want to use are not extensions methods and take only one paramter.
By the way, the methods are not called the same way in expressions.

Common/src/gRPC/FilterRangeExt.cs
```diff
@@ -0,0 +1,153 @@
...
+  /// <param name="filterOperator">The gRPC Enum</param>
+  /// <returns>
+  ///   The <see cref="ExpressionType" /> that represent the operation
+  /// </returns>
+  public static ExpressionType ToExpressionType(this FilterDateOperator filterOperator)
+    => filterOperator switch
+       {
+         FilterDateOperator.Before        => ExpressionType.LessThan,
+         FilterDateOperator.BeforeOrEqual => ExpressionType.LessThanOrEqual,
+         FilterDateOperator.Equal         => ExpressionType.Equal,
```
ngruelaneo, line 61: Dans ce filtre, contrairement au suivant, il n'y a pas le `NotEqual` est-ce voulu?

aneojgurhem: j'ai l'impression oui @esoubiran-aneo

Common/tests/Auth/AuthenticationIntegrationTest.cs
```diff
@@ -37,6 +37,9 @@
 using Armonik.Api.Grpc.V1.SortDirection;
 
 using ArmoniK.Api.gRPC.V1.Submitter;
+
+using Armonik.Api.gRPC.V1.Tasks;
```
ngruelaneo, line 41: il me semble que cela fait doublon avec la ligne 43

ngruelaneo (approved):

Common/src/gRPC/ListApplicationsRequestExt.cs
```diff
@@ -48,29 +57,41 @@ public static class ListApplicationsRequestExt
...
+  ///   Converts gRPC message filters into an <see cref="Expression" /> that represents the filter conditions
+  /// </summary>
+  /// <param name="filters">The gPRC filters</param>
+  /// <returns>
+  ///   The <see cref="Expression" /> that represents the filter conditions
+  /// </returns>
+  /// <exception cref="ArgumentOutOfRangeException">the given message is not recognized</exception>
+  public static Expression<Func<TaskData, bool>> ToApplicationFilter(this Filters filters)
   {
     var predicate = PredicateBuilder.New<TaskData>();
```
melflitty, line 70: the default value given by this function is false?

aneojgurhem: yes

ngruelaneo (approved):

PR #453, opened 2023-07-31
refactor: remove LinqKit

melflitty:
Some functions of LinqKit were used to generate predicate expressions for filters. We don't want to use this package anymore.

### DIFF
--- a/Common/src/gRPC/ExpressionExt.cs
+++ b/Common/src/gRPC/ExpressionExt.cs
@@ -1,29 +1,48 @@
+// This file is part of the ArmoniK project
+// 
+// Copyright (C) ANEO, 2021-2023. All rights reserved.
+// 
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as published
+// by the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+// 
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY, without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+// 
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
 using System;
 using System.Linq.Expressions;
+
 public static class ExpressionExt
 {
   /// <summary>
-  /// Combines two predicate expressions using a logical AND condition
+  ///   Combines two predicate expressions using a logical AND condition
   /// </summary>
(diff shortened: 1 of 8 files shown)


Common/src/gRPC/ExpressionExt.cs
```diff
@@ -0,0 +1,92 @@
...
+    var rightVisitor = new ReplaceExpressionVisitor(expr2.Parameters[0],
+                                                    parameter);
+    var right = rightVisitor.Visit(expr2.Body);
+
+    return Expression.Lambda<Func<T, bool>>(Expression.OrElse(left,
+                                                              right),
+                                            parameter);
+  }
+
+  private class ReplaceExpressionVisitor : ExpressionVisitor
```
aneojgurhem, line 75: Pretty clean solution !
I'd use Expression.MakeBinary to factorize code and make the generic function private.

Common/src/gRPC/ExpressionExt.cs
```diff
@@ -54,8 +55,8 @@ public static class ExpressionExt
   /// <param name="expr1"> The first predicate expression to combine </param>
   /// <param name="expr2"> The second predicate expression to combine </param>
   /// <returns> A new predicate expression that represents the logical OR of the two expressions </returns>
-  public static Expression<Func<T, bool>> ExpressionOr<T>(this Expression<Func<T, bool>> expr1,
-                                                          Expression<Func<T, bool>>      expr2)
+  public static Expression<Func<T, bool>>? ExpressionOr<T>(this Expression<Func<T, bool>> expr1,
```
aneojgurhem, line 58: should  not return null
(you missed the factorization part)

melflitty: I was doing it . will push that part after ensuring the generic function works well

melflitty: Sorry, I misunderstood your comment . Done!

Common/src/gRPC/ExpressionExt.cs
```diff
@@ -0,0 +1,89 @@
...
+                                                   parameter);
+    var left = leftVisitor.Visit(expr1.Body);
+
+    var rightVisitor = new ReplaceExpressionVisitor(expr2.Parameters[0],
+                                                    parameter);
+    var right = rightVisitor.Visit(expr2.Body);
+
+    return Expression.Lambda<Func<T, bool>>(Expression.MakeBinary(expressionType,
+                                                                  left!,
+                                                                  right!),
```
aneojgurhem, line 41: I think you should properly check for null here !

lemaitre-aneo: Personally, I consider `!` a proper null check.

PR #458, opened 2023-08-11
refactor: upload objects into S3 using multipart upload 

melflitty:


Adaptors/S3/src/ObjectStorage.cs
```diff
@@ -213,76 +179,70 @@ await keyList.Chunk(MaxAllowedKeysPerDelete)
                        bucketName_);
```
lemaitre-aneo, line 179:
```suggestion
                       bucketName_);
         return false;
```

Actually, I'm not sure if you should return instead of rethrow.

Adaptors/S3/src/ObjectStorage.cs
```diff
@@ -213,76 +179,70 @@ await keyList.Chunk(MaxAllowedKeysPerDelete)
...
+                       "Multipart upload is being aborted");
+      var abortMpuRequest = new AbortMultipartUploadRequest
+                            {
+                              BucketName = bucketName_,
+                              Key        = $"{objectStorageName_}{key}",
+                              UploadId   = initResponse.UploadId,
+                            };
+      await s3Client_.AbortMultipartUploadAsync(abortMpuRequest,
+                                                cancellationToken);
+    }
```
lemaitre-aneo, line 246:
```suggestion
    catch
    {
      var abortMpuRequest = new AbortMultipartUploadRequest
                            {
                              BucketName = bucketName_,
                              Key        = $"{objectStorageName_}{key}",
                              UploadId   = initResponse.UploadId,
                            };
      await s3Client_.AbortMultipartUploadAsync(abortMpuRequest,
                                                cancellationToken);
      throw;
    }
```

Adaptors/S3/src/ObjectStorage.cs
```diff
@@ -124,7 +99,7 @@ await taskList.WhenAll()
     try
     {
       response = await s3Client_.GetObjectAsync(bucketName_,
-                                                $"{objectStorageName_}{key}_count",
+                                                $"{objectStorageName_}{key}",
```
lemaitre-aneo, line 102: You can store this string interpolation into a variable and reuse it.

Adaptors/S3/src/UploadMultiPartHelper.cs
```diff
@@ -0,0 +1,112 @@
...
+
+namespace ArmoniK.Core.Adapters.S3;
+
+internal class UploadMultiPartHelper
+{
+  private const long MinPartSize  = 5 * 1024 * 1024; // 5 MiB
+  private const int  MaxPartCount = 10000;
+  private const long MaxPartSize  = 5L * 1024 * 1024 * 1024; // 5 GiB
+
+  public static async Task<List<UploadPartRequest>> PreparePartRequestsAsync(string                                 bucketName,
```
lemaitre-aneo, line 28: Instead of returning a Task of list of requests, I think it should be better to return a IAsyncEnumerable of requests.

Adaptors/S3/src/UploadMultiPartHelper.cs
```diff
@@ -0,0 +1,109 @@
...
+
+using Amazon.S3.Model;
+
+namespace ArmoniK.Core.Adapters.S3;
+
+internal class UploadMultiPartHelper
+{
+  private const long MinPartSize  = 5 * 1024 * 1024; // 5 MiB
+  private const int  MaxPartCount = 10000;
+  private const long MaxPartSize  = 5L * 1024 * 1024 * 1024; // 5 GiB
```
aneojgurhem, line 26: Is this L really needed ?

melflitty: Yes

melflitty: without L the 5*1024*1024*1024 will be evaluated with int32 then casted to long. This results in an overflow because the value is > int.MaxValue

Adaptors/S3/src/ObjectStorage.cs
```diff
@@ -134,155 +109,142 @@ await taskList.WhenAll()
...
+                                                                         valueChunks,
+                                                                         cancellationToken);
+      var uploadResponses = await uploadRequest.ParallelSelect(new ParallelTaskOptions(degreeOfParallelism_),
+                                                               async uploadPartRequest =>
+                                                               {
+                                                                 var uploadPartResponse = await s3Client_.UploadPartAsync(uploadPartRequest,
+                                                                                                                          cancellationToken)
+                                                                                                         .ConfigureAwait(false);
+                                                                 return uploadPartResponse;
+                                                               })
```
lemaitre-aneo, line 220:
```suggestion
                                                               uploadPartRequest => s3Client_.UploadPartAsync(uploadPartRequest,
                                                                                             cancellationToken))
```

Adaptors/S3/src/UploadMultiPartHelper.cs
```diff
@@ -0,0 +1,109 @@
...
+
+  public static async IAsyncEnumerable<UploadPartRequest> PreparePartRequestsAsync(string                                 bucketName,
+                                                                                   string                                 objectKey,
+                                                                                   string                                 uploadId,
+                                                                                   IAsyncEnumerable<ReadOnlyMemory<byte>> valueChunks,
+                                                                                   CancellationToken                      cancellationToken)
```
lemaitre-aneo, line 32:
```suggestion
                                                                                   [EnumeratorCancellation] CancellationToken cancellationToken)
```

lemaitre-aneo (dismissed):

Adaptors/S3/src/ObjectStorage.cs
```diff
@@ -55,10 +56,12 @@ public class ObjectStorage : IObjectStorage
                        Options.S3             options,
                        ILogger<ObjectStorage> logger)
   {
-    s3Client_          = s3Client;
-    objectStorageName_ = "objectStorageName";
-    bucketName_        = options.BucketName;
-    logger_            = logger;
+    s3Client_            = s3Client;
+    objectStorageName_   = "objectStorageName";
```
aneojgurhem, line 60: as it is not an input anymore, we can remove it from the xml doc above

Adaptors/S3/src/ObjectStorage.cs
```diff
@@ -89,42 +92,17 @@ public Task<HealthCheckResult> Check(HealthCheckTag tag)
...
-
   /// <inheritdoc />
   public async IAsyncEnumerable<byte[]> GetValuesAsync(string                                     key,
                                                        [EnumeratorCancellation] CancellationToken cancellationToken = default)
   {
-    using var          _        = logger_.LogFunction(objectStorageName_ + key);
-    GetObjectResponse? response = null;
+    var                objectStorageFullName = $"{objectStorageName_}{key}";
+    using var          _                     = logger_.LogFunction(objectStorageFullName);
+    GetObjectResponse? response              = null;
```
aneojgurhem, line 101: This can be removed as it is not used

lemaitre-aneo (dismissed):

aneojgurhem (approved):

PR #464, opened 2023-08-17
feat: create sessions in sessions service

aneojgurhem:


Common/src/gRPC/Services/GrpcSessionsService.cs
```diff
@@ -165,4 +171,45 @@ public class GrpcSessionsService : Sessions.SessionsBase
...
+                         "Partition not found while creating session");
+      throw new RpcException(new Status(StatusCode.InvalidArgument,
+                                        "Partition not found"));
+    }
+    catch (ArmoniKException e)
+    {
+      logger_.LogWarning(e,
+                         "Error while creating session");
+      throw new RpcException(new Status(StatusCode.Internal,
+                                        "Internal Armonik Exception, see Submitter logs"));
```
dbrasseur-aneo, line 205:
```suggestion
                                        "Internal ArmoniK Exception, see Submitter logs"));
```

Common/src/Storage/SessionLifeCycleHelper.cs
```diff
@@ -0,0 +1,96 @@
+// This file is part of the ArmoniK project
+//
+// Copyright (C) ANEO, 2021-$CURRENT_YEAR$. All rights reserved.
```
dbrasseur-aneo, line 3:
```suggestion
// Copyright (C) ANEO, 2021-$YEAR$. All rights reserved.
```
Isn't it ?

dbrasseur-aneo (changes requested):

Common/src/Storage/SessionLifeCycleHelper.cs
aneojgurhem: It is CURRENT_YEAR

lemaitre-aneo: It should be expanded.

aneojgurhem: It will in the next commit

dbrasseur-aneo (approved):

PR #473, opened 2023-08-23
refactor: add missing arm in a switch case

aneojgurhem:


Common/src/gRPC/ListResultsRequestExt.cs
```diff
@@ -39,6 +39,7 @@ public static class ListResultsRequestExt
     => sort.Field.FieldCase switch
        {
          ResultField.FieldOneofCase.ResultRawField => sort.Field.ResultRawField.Field.ToField(),
+         ResultField.FieldOneofCase.None           => throw new ArgumentOutOfRangeException(nameof(sort)),
```
dbrasseur-aneo, line 42: That's already handled by the "_" case isn't it ?

aneojgurhem: yes but resharper wants to see ervery arm in the switch case

dbrasseur-aneo (approved):

PR #478, opened 2023-08-25
refactor: use OTLP exporter instead of zipkin

aneojgurhem:


dbrasseur-aneo (approved):
Out of curiosity, what are the benefits of using OTLP instead of zipkin ?

aneojgurhem:
Zipkin only works for Zipkin whereas  OLTP works for every receiver that implement OpenTelemetry standard

PR #479, opened 2023-08-28
feat: Add filters in event API

aneojgurhem:


Common/tests/TestBase/TaskWatcherTestBase.cs
```diff
@@ -473,4 +473,50 @@ await foreach (var cur in watchEnumerator.WithCancellation(cts.Token)
                       newResults[2]);
     }
   }
+
+  [Test]
+  public async Task WatchTaskStatusUpdateWithComplexFilterShouldSucceed()
+  {
+    if (RunTests)
+    {
```
lemaitre-aneo, line 481:
```suggestion
    if (!RunTests)
    {
      return;
    }
```

aneojgurhem: As it is done like that in every test, it is not the place to change that.

Common/src/gRPC/Validators/EventSubscriptionRequestValidator.cs
```diff
@@ -0,0 +1,124 @@
...
+                case ResultRawEnumField.Unspecified:
+                  context.AddFailure($"Cannot filter {filterField.Field.ResultRawField.Field} on in this API");
+                  break;
+                case ResultRawEnumField.OwnerTaskId:
+                case ResultRawEnumField.SessionId:
+                case ResultRawEnumField.Name:
+                case ResultRawEnumField.ResultId:
+                  break;
+                default:
+                  throw new ArgumentOutOfRangeException();
```
lemaitre-aneo, line 60: `throw` or `context.AddFailure`?

Common/src/gRPC/Validators/EventSubscriptionRequestValidator.cs
```diff
@@ -0,0 +1,124 @@
...
+  /// <inheritdoc />
+  public override ValidationResult Validate(ValidationContext<EventSubscriptionRequest> context)
+  {
+    if (context.InstanceToValidate.ResultsFilters != null)
+    {
+      foreach (var filtersAnd in context.InstanceToValidate.ResultsFilters.Or)
+      {
+        foreach (var filterField in filtersAnd.And)
+        {
+          switch (filterField.Field.FieldCase)
```
lemaitre-aneo, line 43: I think you should check for `filterField.Field` nullness

lemaitre-aneo (dismissed):

Common/src/gRPC/Validators/EventSubscriptionRequestValidator.cs
aneojgurhem: At this point, it is not the request that is not valid but our code so I think the exception is more appropriate.

Common/src/gRPC/Validators/EventSubscriptionRequestValidator.cs
aneojgurhem: Yes!

Common/src/gRPC/Validators/EventSubscriptionRequestValidator.cs
lemaitre-aneo: I think this can happen if the client sends an invalid number for the enum (like 100).

lemaitre-aneo (approved):

lemaitre-aneo (approved):

PR #481, opened 2023-08-29
chore: create terraform tracing module that exports traces to file and zipkin

aneojgurhem:


terraform/modules/tracing/locals.tf
melflitty, line 1: This file is empty. It can be removed.

lemaitre-aneo (approved):

PR #482, opened 2023-09-01
feat: allow for payload submission after task creation

aneojgurhem:


Common/src/gRPC/Services/Submitter.cs
```diff
@@ -541,20 +541,33 @@ await foreach (var taskRequest in taskRequests.WithCancellation(cancellationToke
...
-      payloadUploadTasks.Add(objectStorage_.AddOrUpdateAsync(taskId,
+      payloadUploadTasks.Add(objectStorage_.AddOrUpdateAsync(payloadId,
                                                              taskRequest.PayloadChunks,
                                                              cancellationToken));
     }
 
     await payloadUploadTasks.WhenAll()
                             .ConfigureAwait(false);
 
+    await resultTable_.Create(requests.Select(request => new Result(sessionId,
```
lemaitre-aneo, line 560: What does this means? Do you change the requirement on whether the payload should be created before hand?

ngruelaneo (dismissed):

Common/src/gRPC/Services/GrpcResultsService.cs
```diff
@@ -292,6 +309,19 @@ public class GrpcResultsService : Results.ResultsBase
...
                                          .ConfigureAwait(false);
 
+      await TaskLifeCycleHelper.ResolveDependencies(taskTable_,
+                                                    resultTable_,
+                                                    pushQueueStorage_,
+                                                    id.SessionId,
+                                                    new List<string>
+                                                    {
+                                                      id.ResultId,
+                                                    },
```
lemaitre-aneo, line 319: nitpick:
```suggestion
                                                    new string[]
                                                    {
                                                      id.ResultId,
                                                    },
```

Common/src/gRPC/Services/Submitter.cs
aneojgurhem: The payload is still required to be created in the tasks service.
In the submitter service, payload metadata were not created at all.

lemaitre-aneo: That was nasty. Fair enough.

lemaitre-aneo (approved):

PR #486, opened 2023-09-04
fix: results created from client should belong to the session

aneojgurhem:


Common/src/gRPC/Services/GrpcResultsService.cs
```diff
@@ -156,7 +156,7 @@ public class GrpcResultsService : Results.ResultsBase
                                                                          Guid.NewGuid()
                                                                              .ToString(),
                                                                          rc.Name,
-                                                                         "",
+                                                                         request.SessionId,
```
melflitty, line 159: is it normal that SessionId is passed two times :one for SessionId and another for OwnerTaskId?

aneojgurhem: Yes that is the purpose of this pr

melflitty (approved):

PR #491, opened 2023-09-06
chore: Use bytes type for tosend as log_message and Add logging module

jeremyzyngeraneo:


tools/logs2seq.py
```diff
@@ -43,40 +52,39 @@
 s3 = boto3.client('s3')
 s3.download_file(args.bucket_name, obj_name, file_name)
 
-def process_json_log(url, file_name):
-    batch = 0
+def process_json_log(url, file_name: IO[bytes]):
```
aneojgurhem, line 55:
```suggestion
def process_json_log(url: str, file_name: str):
```

aneojgurhem (changes requested):

tools/logs2seq.py
```diff
@@ -43,40 +52,39 @@
...
-            if batch > 100:
-                requests.post(url, data=tosend)
-                tosend = ""
-                batch = 0
-            batch = batch + 1
-            ctr = ctr + 1
-        print("sent :", ctr)
-        if tosend != "":
+                ctr = ctr + 1
+                log_message = bytes(line, "utf-8")
```
aneojgurhem, line 62:
```suggestion
                log_message = bytes(line + "\n", "utf-8")
```

aneojgurhem (changes requested):
```python
                parsed = json.loads(line)
                if "@t" not in parsed:
                    continue
```

We need to check that there is this field because if the field is not present the data are not accepted. Can you add it ?

jeremyzyngeraneo:
`if line.startswith("{\"@t"):`

Is it possible to filter the data like this ?

aneojgurhem:
No `@t` can be anywhere in the document

aneojgurhem (approved):

PR #492, opened 2023-09-07
refactor: allow the usage of S3 option UseChunkEncoding

aneojgurhem:


melflitty (approved):

Adaptors/S3/src/Options/S3.cs
```diff
@@ -35,4 +35,9 @@ public class S3
   ///   Size of one chunk when downloading an object by chunks
   /// </summary>
   public int ChunkDownloadSize { get; set; } = 65536;
+
+  /// <summary>
+  ///   Use Chunk Encoding during upload
```
lemaitre-aneo, line 40: Maybe add more explanation to what it does.

PR #498, opened 2023-09-18
refactor: logs2seq with class

jeremyzyngeraneo:


tools/logs2seq.py
```diff
@@ -34,73 +34,100 @@
...
                 except JSONDecodeError as e:
                     logger.warning(f"Failed to parse JSON: {e}")
-        logger.info(f"sent : {ctr}")
-        if tosend != b"":
-            requests.post(url, data=tosend)
-
-if file_name.endswith(".json.tar.gz"):
-    process_jsongz_log(args.url, file_name)
-elif file_name.endswith(".json"):
-    process_json_log(args.url, file_name)
```
aneojgurhem, line 106: You've lost that and that is not good !

jeremyzyngeraneo: I tried to handle file extensions with the read_file method  for better readability.
Isn't it possible?

PR #528, opened 2023-11-07
fix: do not cancel start of task in database in pollster to execute ready task before grace delay cancels it

### LINKED ISSUE #527: Failing to acquire a task in processing without OwnerPodId leads to this task entering acquisition deadlock
A task was cancelled due to pod being destroyed. Then, task execution grace delay period comes in before cancelling the task. However, this delay was not properly taken into account leading to the cancellation of what seems to be the request to change task status to Processing (StartTask). However, status seems to have been properly updated to Processing. Then, ReleaseTask happens almost at the same time. What's strange is that status is not updated to Submitted but OwnerPodId is removed.

Then agent tries to acquire the task which is in Processing status according to logs. Acquisition does not succeed since it requires task to be in the Submitted status. Ownerpodid is still empty so task acquisition does not succeed and task is postponed.

Version : 0.13.2

### LINKED ISSUE #530: change condition for start task (use == status.dispatched instead of != list())

aneojgurhem:
fix #527 and fix #530


Adaptors/MongoDB/src/TaskTable.cs
```diff
@@ -229,7 +229,7 @@ public class TaskTable : ITaskTable
           throw new TaskNotFoundException($"Task {taskId} not found");
         }
 
-        throw new ArmoniKException($"Task already in a terminal state - {taskStatus.Single()} to {TaskStatus.Processing}");
+        throw new ArmoniKException($"Task was not acquired - {taskStatus.Single()} to {TaskStatus.Processing}");
```
lemaitre-aneo, line 232: This is not AcquireTask but StartTask.

lemaitre-aneo (approved):

Adaptors/MongoDB/src/TaskTable.cs
aneojgurhem: But the root cause was that it was not acquired

PR #540, opened 2023-11-21
refactor: UpdateManyResults

lemaitre-aneo:
Implement `UpdateManyResults` and use it to implement some `ResultTable` methods.

### DIFF
--- a/Adaptors/Memory/src/ResultTable.cs
+++ b/Adaptors/Memory/src/ResultTable.cs
@@ -260,10 +260,14 @@ public class ResultTable : IResultTable
       throw new SessionNotFoundException($"Session '{sessionId}' not found");
     }
 
-    return Task.FromResult(session.AddOrUpdate(resultId,
-                                               _ => throw new ResultNotFoundException($"Result '{resultId}' not found"),
-                                               (_,
-                                                data) => new Result(data,
-                                                                    updates)));
+    if (!session.TryGetValue(resultId,
+                             out var result))
+    {
+      throw new ResultNotFoundException($"Result '{resultId}' not found");
+    }
+
+    session[resultId] = new Result(result,
+                                   updates);
+    return Task.FromResult(result);
   }
 }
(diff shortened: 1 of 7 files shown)


Adaptors/Memory/src/ResultTable.cs
```diff
@@ -308,4 +223,47 @@ public Task<HealthCheckResult> Check(HealthCheckTag tag)
...
+                                      ICollection<(Expression<Func<Result, object?>> selector, object? newValue)> updates,
+                                      CancellationToken                                                           cancellationToken = default)
+  {
+    if (!results_.TryGetValue(sessionId,
+                              out var session))
+    {
+      throw new SessionNotFoundException($"Session '{sessionId}' not found");
+    }
+
+    return Task.FromResult(session.AddOrUpdate(resultId,
```
aneojgurhem, line 263: It should be result before update. (It is what's explained in ITaskTable doc). Maybe it returns the value before update ?
If this is based on what's done in TaskTable implementation, there may be an issue and a missing test

Adaptors/MongoDB/src/ResultTable.cs
```diff
@@ -379,6 +273,66 @@ public class ResultTable : IResultTable
                           .ConfigureAwait(false);
   }
 
+  /// <inheritdoc />
+  public async Task<Result> UpdateOneResult(string                                                                      sessionId,
```
aneojgurhem, line 277: I don't know if it is time we remove it from the interfaces.
Maybe in another PR

aneojgurhem (commented):
Overall, I'm okay with the dev. Thanks !

Adaptors/MongoDB/src/ResultTable.cs
lemaitre-aneo: Yes, I think it is better in another PR.

Adaptors/Memory/src/ResultTable.cs
lemaitre-aneo: Yes, it is based on the TaskTable implementation. I will fix both.

aneojgurhem (approved):

PR #552, opened 2023-12-12
feat: Add size to the results

lemaitre-aneo:


Adaptors/S3/src/ObjectStorage.cs
```diff
@@ -331,5 +341,7 @@ await foreach (var chunk in valueChunks.WithCancellation(cancellationToken)
                         };
       yield return partRequest;
     }
+
+    sizeBox[0] = size;
```
melflitty, line 345: I am not sure to understand why we have an array of sizes, but only the first value is used

lemaitre-aneo: It is because you cannot have a ref or out var for an iterator, so I need to box my int so that the iterator can modify the value and "pass it back" to the caller. There are multiple ways to do it, and I think the array with a single element is one of the simplest. I will add a comment about it.

melflitty (dismissed):

aneojgurhem (commented):
filtering and ordering results on size does not seem to be added. Was it added in APIs ?

aneojgurhem (approved):

PR #555, opened 2023-12-13
fix: reduce metrics exported

lemaitre-aneo:
Revamp how the metrics are computed and exported.

This enables to filter out some metrics that are not useful in order to reduce the load on the database.
It also improves the database query to avoid scanning all documents, only the indexed keys are scanned now.

For a database with 1.5M tasks, here are the performance to count all the tasks:
- Before: 1000ms
- After, without proper index: 2000ms
- After, with proper (combined) index: 450ms

### DIFF
--- a/Control/Metrics/src/ArmoniKMeter.cs
+++ b/Control/Metrics/src/ArmoniKMeter.cs
@@ -24,6 +24,7 @@ using System.Threading.Tasks;
 
 using ArmoniK.Api.Common.Utils;
 using ArmoniK.Core.Common.Storage;
+using ArmoniK.Core.Control.Metrics.Options;
 using ArmoniK.Utils;
 
 using Microsoft.Extensions.Hosting;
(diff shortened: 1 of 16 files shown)


Control/Metrics/src/ArmoniKMeter.cs
```diff
@@ -35,155 +35,184 @@ namespace ArmoniK.Core.Control.Metrics;
...
+    TaskStatus.Dispatched,
+    //TaskStatus.Completed,
+    //TaskStatus.Error,
+    //TaskStatus.Timeout,
+    //TaskStatus.Cancelling,
+    //TaskStatus.Cancelled,
+    TaskStatus.Processing,
+    TaskStatus.Processed,
+    TaskStatus.Retried,
+  };
```
ngruelaneo, line 54: will it be possible to leave the choice at the user and having a configuration file or whatever which will provide the status that will be use in the metrics exporter?

lemaitre-aneo: Yes, I've now added the code for it.

aneojgurhem (dismissed):

aneojgurhem (approved):

PR #559, opened 2023-12-18
feat: implement Health Check Service

aneojgurhem:


aneojgurhem:
> There already exist a `GrpcHealthCheckService`, but it does seem to be used. I think you should remove it. Or you can replace it by your new one.

The existing one is the official health check that can be requested through grpc. It could be used by Kube when the support for grpc health will be generally available.

Common/tests/GrpcHealthChecksServiceTest.cs
```diff
@@ -0,0 +1,51 @@
...
+
+[TestFixture]
+public class GrpcHealthChecksServiceTest
+{
+  [Test]
+  public async Task CheckHealthShouldSucceed()
+  {
+    var taskTable     = new SimpleTaskTable();
+    var objectStorage = new SimpleObjectStorage();
+    var queueStorage  = new SimplePushQueueStorage();
```
lemaitre-aneo, line 36: I think it would be better to have `Mock<>` and ensure that `Check` is actually called on those mocks.

lemaitre-aneo (approved):

PR #570, opened 2024-01-03
feat: use .Net8.0

aneojgurhem:

PR #572, opened 2024-01-05
feat: use dotnet 8 azurelinux3.0 distroless docker images

aneojgurhem:
# Motivation

Reduce the size of the images and reduce surface of attack.

# Description

Use `mcr.microsoft.com/dotnet/aspnet:8.0-azurelinux3.0-distroless` as base image instead of `mcr.microsoft.com/dotnet/aspnet:8.0`. It mainly changes the way the images are constructed:
- Images are minimalist so it is not possible to create new users. We use `$APP_UID` user defined by the image with ID 1654 as of now instead of `armonikuser` with ID 5000.
- The `/cache` folder is not created anymore. Driver options are given to create the volume underneath and make it properly available for the agent and the worker.

The size of the images is reduced by around 100Mo.

# Testing

- Pipelines in this repo are passing.
- Tests in ArmoniK infrastructure on localhost and in AWS

# Impact

- Reduced surface of attack and less CVEs
- Smaller images
- For compatibility with other workers (extensions, samples), the infrastructure will have to set a common user between the agent and the worker to share the sockets used for communication between the two.

# Additional Information

- https://stackoverflow.com/questions/51456621/docker-volumes-specifying-permissions-using-mount-options
- https://www.howtogeek.com/812961/umask-linux/

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM mcr.microsoft.com/dotnet/aspnet:8.0-jammy-chiseled AS base-linux
+FROM mcr.microsoft.com/dotnet/aspnet:8.0-azurelinux3.0-distroless AS base-linux
 ARG TARGETARCH
 ADD --chmod=755 https://github.com/krallin/tini/releases/download/v0.19.0/tini-static-${TARGETARCH} /tini
 USER $APP_UID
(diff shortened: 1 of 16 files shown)


aneojgurhem:
Need to be rebased when #570 is merged

PR #579, opened 2024-01-11
feat: Implement closing APIs for session service

aneojgurhem:


Common/src/Storage/ISessionTable.cs
```diff
@@ -65,25 +65,14 @@ public interface ISessionTable : IInitializable
...
-  ///   Delete a session
+  ///   Remove a session
   /// </summary>
   /// <param name="sessionId">Id of the session to delete</param>
   /// <param name="cancellationToken">Token used to cancel the execution of the method</param>
   /// <returns>
   ///   Task representing the asynchronous execution of the method
   /// </returns>
-  Task DeleteSessionAsync(string            sessionId,
+  Task RemoveSessionAsync(string            sessionId,
```
lemaitre-aneo, line 75: I am not sure why you renamed this function.

Common/src/Storage/SessionTableExtensions.cs
```diff
@@ -110,4 +112,169 @@ public static class SessionTableExtensions
...
+  /// <param name="cancellationToken">Token used to cancel the execution of the method</param>
+  /// <returns>
+  ///   The metadata of the cancelled session
+  /// </returns>
+  public static async Task<SessionData> CancelSessionAsync(this ISessionTable sessionTable,
+                                                           string             sessionId,
+                                                           CancellationToken  cancellationToken = default)
+    => await sessionTable.UpdateOneSessionAsync(sessionId,
+                                                data => data.Status == SessionStatus.Running || data.Status == SessionStatus.Paused ||
+                                                        data.Status != SessionStatus.Cancelled,
```
lemaitre-aneo, line 130: Your condition is equivalent to the following: 
```c#
data => data.Status != SessionStatus.Cancelled
```

So I think it is not what you want.
```suggestion
                                                data => data.Status == SessionStatus.Running || data.Status == SessionStatus.Paused,
```

Common/src/Storage/SessionTableExtensions.cs
```diff
@@ -110,4 +112,169 @@ public static class SessionTableExtensions
...
+  /// <param name="sessionId">Id of the session to purge</param>
+  /// <param name="cancellationToken">Token used to cancel the execution of the method</param>
+  /// <returns>
+  ///   The metadata of the purged session
+  /// </returns>
+  public static async Task<SessionData> PurgeSessionAsync(this ISessionTable sessionTable,
+                                                          string             sessionId,
+                                                          CancellationToken  cancellationToken = default)
+    => await sessionTable.UpdateOneSessionAsync(sessionId,
+                                                data => data.Status == SessionStatus.Running || data.Status == SessionStatus.Paused,
```
lemaitre-aneo, line 199: I think you should be able to purge a Cancelled session.

Common/src/Storage/SessionTableExtensions.cs
```diff
@@ -110,4 +112,169 @@ public static class SessionTableExtensions
...
+  /// <summary>
+  ///   Cancel a session
+  /// </summary>
+  /// <param name="sessionTable">Interface to manage sessions lifecycle</param>
+  /// <param name="sessionId">Id of the session to cancel</param>
+  /// <param name="cancellationToken">Token used to cancel the execution of the method</param>
+  /// <returns>
+  ///   The metadata of the cancelled session
+  /// </returns>
+  public static async Task<SessionData> CancelSessionAsync(this ISessionTable sessionTable,
```
lemaitre-aneo, line 125: I think this should throw if the session is purged.

Common/src/gRPC/Services/GrpcSessionsService.cs
lemaitre-aneo, line 1: We really need interceptors. That's some heavy boilerplate to handle exceptions we have here...

Common/src/Storage/SessionTableExtensions.cs
```diff
@@ -110,4 +112,169 @@ public static class SessionTableExtensions
...
+  public static async Task<SessionData> PurgeSessionAsync(this ISessionTable sessionTable,
+                                                          string             sessionId,
+                                                          CancellationToken  cancellationToken = default)
+    => await sessionTable.UpdateOneSessionAsync(sessionId,
+                                                data => data.Status == SessionStatus.Running || data.Status == SessionStatus.Paused,
+                                                new List<(Expression<Func<SessionData, object?>> selector, object? newValue)>
+                                                {
+                                                  (model => model.Status, SessionStatus.Purged),
+                                                  (model => model.PurgeDate, DateTime.UtcNow),
+                                                },
```
lemaitre-aneo, line 204: I think Purge should also disable submission.
```suggestion
                                                {
                                                  (model => model.Status, SessionStatus.Purged),
                                                  (model => model.PurgeDate, DateTime.UtcNow),
                                                  (model => model.ClientSubmission, false),
                                                  (model => model.WorkerSubmission, false),
                                                },
```

Common/src/Storage/SessionTableExtensions.cs
```diff
@@ -110,4 +112,169 @@ public static class SessionTableExtensions
...
+  /// <param name="sessionTable">Interface to manage sessions lifecycle</param>
+  /// <param name="sessionId">Id of the session to delete</param>
+  /// <param name="cancellationToken">Token used to cancel the execution of the method</param>
+  /// <returns>
+  ///   The metadata of the deleted session
+  /// </returns>
+  /// <remarks>
+  ///   A session cannot be deleted twice
+  /// </remarks>
+  public static async Task<SessionData> DeleteSessionAsync(this ISessionTable sessionTable,
```
lemaitre-aneo, line 221: Should Delete also purge the session? Would we need the Delete call to also stop submission?

Common/src/Storage/ISessionTable.cs
aneojgurhem: Because it conflicts with the extension method I created.
We have two "delete":
- the extension that sets the status, the date and the TTL
- the one that directly deletes the entry

I think I will remove the last one but I'm still hesitating on whether we should use TTL or synchrone calls to delete entries in the database

Common/src/Storage/SessionTableExtensions.cs
aneojgurhem: That's already the case.

Common/src/gRPC/Services/GrpcSessionsService.cs
aneojgurhem: Yeah I completly agree. I could write some kind of function that check which kind of exception we have but that's a bit more work

Common/src/Storage/SessionTableExtensions.cs
aneojgurhem: That's equivalent to 

```csharp
data => data.Status != SessionStatus.Cancelled
```

only when there is the three statuses.

I don't want to cancel purged or deleted sessions.

Common/src/Storage/SessionTableExtensions.cs
aneojgurhem: We may want to cancel running tasks beforehand

Common/src/Storage/SessionTableExtensions.cs
lemaitre-aneo: > I don't want to cancel purged or deleted sessions.

That's my point. Currently, you are able to cancel purged and deleted sessions. Your condition is all ORs, so `data.Status == SessionStatus.Paused` is included in `data.Status != SessionStatus.Cancelled`

aneojgurhem: Yeah! You are right. I changed the conditions and added some tests.

Common/tests/TestBase/SessionTableTestBase.cs
```diff
@@ -644,4 +644,66 @@ public async Task StopBothSubmissionShouldSucceed()
...
+                                           CancellationToken.None)
+                        .ConfigureAwait(false);
+
+      var session = await SessionTable.GetSessionAsync(sessionId)
+                                      .ConfigureAwait(false);
+
+      Assert.AreEqual(SessionStatus.Purged,
+                      session.Status);
+
+      Assert.ThrowsAsync<SessionNotFoundException>(async () =>
```
lemaitre-aneo, line 670: `SessionNotFoundException` is most likely not the right exception for that.

Common/src/Storage/ISessionTable.cs
```diff
@@ -65,18 +65,7 @@ public interface ISessionTable : IInitializable
...
-  /// <param name="cancellationToken">Token used to cancel the execution of the method</param>
-  /// <returns>
-  ///   The metadata of the cancelled session
-  /// </returns>
-  Task<SessionData> CancelSessionAsync(string            sessionId,
-                                       CancellationToken cancellationToken = default);
-
-  /// <summary>
-  ///   Delete a session
+  ///   Remove a session
```
lemaitre-aneo, line 68: You can keep "delete"

Common/src/Storage/SessionTableExtensions.cs
```diff
@@ -36,6 +39,7 @@ public static class SessionTableExtensions
   /// <returns>
   ///   Data of the session
   /// </returns>
+  /// <exception cref="SessionNotFoundException">if session was not found</exception>
```
lemaitre-aneo, line 42:
```suggestion
  /// <exception cref="SessionNotFoundException">if session was not found or deleted</exception>
```

lemaitre-aneo (approved):

Common/src/Storage/SessionTableExtensions.cs
aneojgurhem: This one is really not found only

lemaitre-aneo (approved):

PR #588, opened 2024-01-18
fix: during acquisition, a condition was always true

aneojgurhem:


ngruelaneo (dismissed):

Common/src/Pollster/TaskHandler.cs
```diff
@@ -174,10 +174,11 @@ public async Task StopCancelledTask()
   ///   Acquisition of the task in the message given to the constructor
   /// </summary>
   /// <returns>
-  ///   Bool representing whether the task has been acquired
+  ///   Integer representing whether the task has been acquired
+  ///   Acquired when return is 0
   /// </returns>
   /// <exception cref="ArgumentException">status of the task is not recognized</exception>
-  public async Task<bool> AcquireTask()
+  public async Task<int> AcquireTask()
```
lemaitre-aneo, line 181: I think it would be better to return an enum instead of an int, but I'm fine with these.

lemaitre-aneo (approved):

Common/src/Pollster/TaskHandler.cs
aneojgurhem: I agree. We need to refactor this function at larger scale but now is not the time

lemaitre-aneo (approved):

PR #592, opened 2024-01-19
test: increase time before triggering cancellation token in MultipleWatchShouldSucceed because test fails on small machines

aneojgurhem:


aneojgurhem:
I wonder if it would be better to remove `TestCase(6)` instead as it does not add much informations

ngruelaneo (approved):

PR #597, opened 2024-01-24
refactor: replace receiver when issue in AMQP plugin

aneojgurhem:


Adaptors/Amqp/src/PullQueueStorage.cs
```diff
@@ -123,9 +112,57 @@ await ConnectionAmqp.Init(cancellationToken)
...
+            if (retry < Options.MaxRetries - 1)
+            {
+              try
+              {
+                await Task.Delay(retry * retry * baseDelay_,
+                                 cancellationToken)
+                          .ConfigureAwait(false);
+
+                var session = new Session(ConnectionAmqp.Connection);
+                receivers_[i] = CreateReceiver(session,
```
lemaitre-aneo, line 137: If you create the session in CreateReceiver and await the receiver within the first try, it would avoid the need for two nested tries altogether.

aneojgurhem: Errors happen during Session creation so you still need the other try catch

lemaitre-aneo (approved):

Common/src/Pollster/TaskHandler.cs
```diff
@@ -131,8 +131,17 @@ public async ValueTask DisposeAsync()
...
-                         .ConfigureAwait(false);
+    try
+    {
+      await messageHandler_.DisposeAsync()
+                           .ConfigureAwait(false);
+    }
+    catch (Exception e)
+    {
+      logger_.LogWarning(e,
+                         "Error while disposing message handler");
```
lemaitre-aneo, line 142:
```suggestion
      logger_.LogWarning(e,
                         "Error while disposing message handler {MessageHandler}. It will appear duplicated",
                         messageHandler_.MessageId);
```

lemaitre-aneo (approved):

PR #605, opened 2024-01-30
fix: recreate AMQP connection when dropped

aneojgurhem:

PR #606, opened 2024-01-30
refactor: release instead of requeue in AMQP plugin

aneojgurhem:


aneojgurhem:
Not good enough as we still have issues with AMQP connection. #605 seems to be better at solving the issues with AMQP Plugin with rabbitMQ.

PR #609, opened 2024-01-30
refactor: Dispose in Pollster

lemaitre-aneo:


Common/src/Pollster/Pollster.cs
```diff
@@ -269,6 +276,22 @@ await cts.CancelAsync()
...
+                                                           // This catch is used to properly dispose every message when the pollster should stop due to
+                                                           // too many errors or fatal error (broken worker, for instance)
+                                                           // Messages may not be all disposed in case the DisposeAsync method throws an exception
+                                                           // whereas it should not (good practices encourage to avoid this behavior)
+                                                           logger_.LogDebug("Start of queue messages disposition");
+
+                                                           while (await messages.MoveNextAsync()
+                                                                                .ConfigureAwait(false))
+                                                           {
+                                                             await messages.Current.DisposeAsync()
```
aneojgurhem, line 290: maybe we need a try catch around that! With plugins, we are not sure what may happen in this function!

lemaitre-aneo: I'm not sure we should guard against that kind of error. If a plugin is not implemented correctly, it is there fault.

lemaitre-aneo: We guarded this in the TaskHandler dispose, so better do it here also.

aneojgurhem (dismissed):

aneojgurhem (approved):

PR #612, opened 2024-02-01
refactor: use cache for push in AMQP plugin

aneojgurhem:


Adaptors/Amqp/src/PushQueueStorage.cs
```diff
@@ -96,37 +94,43 @@ public class PushQueueStorage : QueueStorage, IPushQueueStorage
...
+                                                                                                                                                      .ConfigureAwait(false)),
+                                                                                                                                    Guid.NewGuid()
+                                                                                                                                        .ToString(),
+                                                                                                                                    s),
+                                                                                                      (link,
+                                                                                                       _) => new ValueTask<bool>(!link.IsClosed &&
+                                                                                                                                 !link.Session.IsClosed)));
```
lemaitre-aneo, line 117: This should most likely be a private method. It it a bit hard to read now.

Adaptors/Amqp/src/PushQueueStorage.cs
```diff
@@ -96,37 +94,43 @@ public class PushQueueStorage : QueueStorage, IPushQueueStorage
...
-                                                                                 MessageId = Guid.NewGuid()
-                                                                                                 .ToString(),
-                                                                               },
-                                                                })
-                                                     .ConfigureAwait(false);
+                                         var address = $"{partitionId}###q{whichQueue}";
+
+                                         var pool = senders_.GetOrAdd(address,
+                                                                      s => new ObjectPool<SenderLink>(parallelismLimit_,
+                                                                                                      async token => new SenderLink(new Session(await ConnectionAmqp
```
lemaitre-aneo, line 109: So each Link have its own Session. Is it something we want to keep?

Adaptors/QueueCommon/src/Amqp.cs
```diff
@@ -36,4 +36,5 @@ public class Amqp
   public bool   AllowHostMismatch { get; set; }
   public int    MaxRetries        { get; set; }
   public int    LinkCredit        { get; set; }
+  public int    ParallelismLimit  { get; set; } = 200;
```
lemaitre-aneo, line 39: That is very high. I don't think it is a good idea. You can put 0 to use the number of threads by default.

Adaptors/Amqp/src/PushQueueStorage.cs
```diff
@@ -96,37 +94,43 @@ public class PushQueueStorage : QueueStorage, IPushQueueStorage
...
+                                                                                                                                                      .ConfigureAwait(false)),
+                                                                                                                                    Guid.NewGuid()
+                                                                                                                                        .ToString(),
+                                                                                                                                    s),
+                                                                                                      (link,
+                                                                                                       _) => new ValueTask<bool>(!link.IsClosed &&
+                                                                                                                                 !link.Session.IsClosed)));
+
+                                         await pool.WithInstanceAsync(sender => sender.SendAsync(new Message(Encoding.UTF8.GetBytes(msgData.TaskId))
```
lemaitre-aneo, line 119: You did not pass the cancellation token to SendAsync.

lemaitre-aneo (dismissed):

Adaptors/Amqp/src/PushQueueStorage.cs
aneojgurhem: for now yes

Adaptors/Amqp/src/PushQueueStorage.cs
aneojgurhem: It does not take cancellation tokens

lemaitre-aneo (approved):

PR #615, opened 2024-02-02
fix: stop failed polling agents and let them be restarted by docker

aneojgurhem:


lemaitre-aneo (dismissed):
Does "StopApplication" exit with a non-zero return code ? If no, the restart on-failure will not work.

lemaitre-aneo (approved):

PR #618, opened 2024-02-05
chore: Remove warnings

lemaitre-aneo:


Control/Metrics/src/Program.cs
```diff
@@ -91,22 +91,18 @@ public static async Task<int> Main(string[] args)
...
-                                                     Predicate = check => check.Tags.Contains(nameof(HealthCheckTag.Liveness)),
-                                                   });
-                         endpoints.MapControllers();
-                       });
+      app.MapHealthChecks("/liveness",
+                          new HealthCheckOptions
+                          {
+                            Predicate = check => check.Tags.Contains(nameof(HealthCheckTag.Liveness)),
+                          });
+      app.MapControllers();
```
aneojgurhem, line 105: why is it added here and not in the polling agent ?

Tests/Stream/Client/TaskSubmissionTests.cs
```diff
@@ -65,7 +67,7 @@ public void SetUp()
                                             .AddEnvironmentVariables();
     var configuration = builder.Build();
     var options = configuration.GetRequiredSection(GrpcClient.SettingSection)
```
aneojgurhem, line 69: we have an extension for that !

Tests/Stream/Client/StreamWrapperTests.cs
```diff
@@ -60,7 +61,7 @@ public void SetUp()
                                             .AddEnvironmentVariables();
     var configuration = builder.Build();
     var options = configuration.GetRequiredSection(GrpcClient.SettingSection)
-                               .Get<GrpcClient>();
```
aneojgurhem, line 63: we have an extension to avoid that

aneojgurhem (approved):

PR #619, opened 2024-02-05
refactor: UpdateDefinition

lemaitre-aneo:


Common/src/Storage/TaskTableExtensions.cs
```diff
@@ -509,18 +522,20 @@ public static class TaskTableExtensions
...
+                                                                                 taskData.ReceptionDate)
+                                                                            .Set(tdm => tdm.AcquisitionDate,
+                                                                                 taskData.AcquisitionDate)
+                                                                            .Set(tdm => tdm.Status,
+                                                                                 TaskStatus.Dispatched),
                                             cancellationToken: cancellationToken)
                              .ConfigureAwait(false);
 
-    taskTable.Logger.LogInformation("Acquire task {task} on {podName} with {success}",
+    taskTable.Logger.LogInformation("Acquire task {Task} on {PodName} with {Success}",
```
aneojgurhem, line 538: why changing that ? It will change the filters on logs

lemaitre-aneo: I think we should transform all logs to use capitals, but I agree it is not the right PR to do it.

Adaptors/MongoDB/src/ResultTable.cs
```diff
@@ -333,6 +334,39 @@ public class ResultTable : IResultTable
...
+
+    var updateResult = await resultCollection.BulkWriteAsync(requests,
+                                                             new BulkWriteOptions
+                                                             {
+                                                               IsOrdered = false,
+                                                             },
+                                                             cancellationToken)
+                                             .ConfigureAwait(false);
+
+    return updateResult.ModifiedCount;
```
aneojgurhem, line 367: In `UpdateManyResults` we return MatchedCount. We may want to stay coherent

aneojgurhem (approved):

PR #621, opened 2024-02-06
fix: dispose CancellationTokenSource in PubSub Heart

aneojgurhem:


Adaptors/PubSub/src/Heart.cs
```diff
@@ -57,6 +57,7 @@ public class Heart
   public async Task Stop()
   {
     stoppedHeartCts_?.Cancel();
+    stoppedHeartCts_?.Dispose();
```
lemaitre-aneo, line 60: It will make the runningTask fail. I don't know if it is problematic or not.

PR #622, opened 2024-02-06
feat: close submission in session

aneojgurhem:


ngruelaneo (approved):

Common/src/Exceptions/SubmissionClosed.cs
```diff
@@ -0,0 +1,40 @@
...
+// 
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+using System;
+
+namespace ArmoniK.Core.Common.Exceptions;
+
+[Serializable]
+public class SubmissionClosed : ArmoniKException
```
lemaitre-aneo, line 23:
```suggestion
public class SubmissionClosedException : ArmoniKException
```

Common/tests/Pollster/AgentTest.cs
```diff
@@ -678,6 +682,43 @@ await holder.Agent.FinalizeTaskCreation(CancellationToken.None)
...
+                          holder.QueueStorage,
+                          holder.ResultTable,
+                          holder.TaskTable,
+                          sessionData,
+                          holder.TaskData,
+                          holder.Folder,
+                          holder.Token,
+                          NullLogger.Instance);
+
+    Assert.ThrowsAsync<SubmissionClosed>(() => agent.SubmitTasks(new TaskSubmissionRequest[]
```
lemaitre-aneo, line 709: `Assert.That` ?

lemaitre-aneo (commented):
Is it expected to catch SubmissionClosed and transform it in RpcException only in TaskService, but not in AgentService?

Common/tests/Pollster/AgentTest.cs
aneojgurhem: I'll do it with the others!

lemaitre-aneo (approved):

PR #631, opened 2024-02-08
feat: pause and resume session

aneojgurhem:


Common/src/Storage/TaskLifeCycleHelper.cs
```diff
@@ -469,4 +469,31 @@ await taskTable.FinalizeTaskCreation(messages.Select(task => task.TaskId)
...
+                                            .GroupBy(msg => (msg.Options.PartitionId, msg.Options.Priority))
+                                            .WithCancellation(cancellationToken)
+                                            .ConfigureAwait(false))
+    {
+      await pushQueueStorage.PushMessagesAsync(await grouping.ToListAsync(cancellationToken)
+                                                             .ConfigureAwait(false),
+                                               grouping.Key.PartitionId,
+                                               cancellationToken)
+                            .ConfigureAwait(false);
+    }
```
lemaitre-aneo, line 493: I think this code already exist somewhere. Maybe in the agent, or in FinalizeTaskCreation

Common/src/Storage/TaskLifeCycleHelper.cs
```diff
@@ -469,4 +469,31 @@ await taskTable.FinalizeTaskCreation(messages.Select(task => task.TaskId)
...
+
+    return await sessionTable.ResumeSessionAsync(sessionId,
+                                                 cancellationToken)
+                             .ConfigureAwait(false);
```
lemaitre-aneo, line 497: The database must be updated before enqueueing the tasks. Other some tasks may be popped from the queue and not executing because the session is still paused.

Common/tests/Helpers/SimplePullQueueStorageChannel.cs
```diff
@@ -30,7 +30,7 @@
 
 namespace ArmoniK.Core.Common.Tests.Helpers;
 
-public class SimplePullQueueStorageChannel : IPullQueueStorage
+public class SimplePullQueueStorageChannel : IPullQueueStorage, IPushQueueStorage
```
lemaitre-aneo, line 33: This looks more and more like a Memory Queue Adaptor. MAybe we should do that instead.
Most likely in another PR.

Common/tests/Pollster/TaskHandlerTest.cs
```diff
@@ -964,6 +964,43 @@ public async Task AcquireTaskFromCancelledSessionShouldFail()
                                        .TaskId);
   }
 
+  [Test]
+  public async Task AcquireTaskFromPausedSessionShouldFail()
```
lemaitre-aneo, line 968: I thought you had a framework to test the acquisition. Can't you use it for this test?

aneojgurhem: It does not manage session data so I did like the other test on the subject

Common/src/Storage/TaskLifeCycleHelper.cs
aneojgurhem: Yeah but not quite. It is the function above. It also FinalizeTasKCreation which we do not want to do here

aneojgurhem: It is also async whereas the other is not

PR #632, opened 2024-02-09
refactor: only compute throughput for htcmock and bench

aneojgurhem:


Tests/Common/Client/src/GrpcChannelExt.cs
```diff
@@ -428,4 +467,81 @@ await foreach (var resp in streamingCall.ResponseStream.ReadAllAsync(cancellatio
...
+    var count           = end.Total;
+    var sessionDuration = (last.EndedAt - first.CreatedAt).ToTimeSpan();
+
+    logger.LogInformation("Throughput for session {session} : {sessionThroughput} task/s ({nTasks} tasks in {timespan})",
+                          sessionId,
+                          count / sessionDuration.TotalMilliseconds * 1000,
+                          count,
+                          sessionDuration);
+
+    return (long)(sessionDuration.TotalMilliseconds * 1000 / count);
```
lemaitre-aneo, line 545: If you print the throughtput in this function, it can return void.

Tests/Common/Client/src/GrpcChannelExt.cs
```diff
@@ -428,4 +467,81 @@ await foreach (var resp in streamingCall.ResponseStream.ReadAllAsync(cancellatio
       }
     }
   }
+
+
+  public static async Task<long> ComputeThroughput(this ChannelBase  channel,
```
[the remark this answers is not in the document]
aneojgurhem: Seemed a good idea. What do you suggest ?

Tests/Common/Client/src/GrpcChannelExt.cs
aneojgurhem: I thought that if someone need to present  the data in another way, it would be good to return the info.

Tests/Common/Client/src/GrpcChannelExt.cs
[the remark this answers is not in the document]
lemaitre-aneo: double

Tests/Common/Client/src/GrpcChannelExt.cs
lemaitre-aneo: Then, I would recommend to put the logs outside

lemaitre-aneo (approved):

lemaitre-aneo (approved):

PR #634, opened 2024-02-09
feat: delete session

aneojgurhem:


lemaitre-aneo (approved):

Common/src/Storage/ITaskTable.cs
```diff
@@ -119,6 +119,17 @@ public interface ITaskTable : IInitializable
...
 
+  /// <summary>
+  ///   Remove tasks from the data base given their session id
+  /// </summary>
+  /// <param name="sessionId">Id of the session from which tasks should be deleted</param>
+  /// <param name="cancellationToken">Token used to cancel the execution of the method</param>
+  /// <returns>
+  ///   Task representing the asynchronous execution of the method
+  /// </returns>
+  Task DeleteTasksAsync(string            sessionId,
```
lemaitre-aneo, line 130: Maybe just a filter?

lemaitre-aneo (approved):

Common/src/Storage/ITaskTable.cs
aneojgurhem: If the need arises, we can refactor it

PR #636, opened 2024-02-13
fix: put tasks that stop worker as error

aneojgurhem:


ngruelaneo (dismissed):

lemaitre-aneo (approved):

lemaitre-aneo:
Just a remark, you called your worker segfault, but it just exists without segfaulting. Might not be the best name. Maybe CrashingWorker would be better.

ngruelaneo (approved):

PR #649, opened 2024-02-20
fix: Activities work properly with pipelining

aneojgurhem:


Common/src/Pollster/TaskHandler.cs
```diff
@@ -214,6 +248,14 @@ public async Task<AcquisitionStatus> AcquireTask()
                                                  CancellationToken.None)
                                   .ConfigureAwait(false);
 
+      activity_?.SetTag("SessionId",
+                        taskData_.SessionId);
+      activity_?.SetBaggage("SessionId",
+                            taskData_.SessionId);
```
lemaitre-aneo, line 254: You could add an extension to have a `SetTagAndBaggage`

lemaitre-aneo (approved):

PR #653, opened 2024-02-27
feat: expose telemetry metrics

aneojgurhem:
- https://github.com/prometheus/prometheus/blob/release-2.50/documentation/examples/prometheus-docker.yml
- https://prometheus.io/docs/prometheus/latest/configuration/configuration/#docker_sd_config
- https://stackoverflow.com/questions/76084366/adding-global-tags-with-system-diagnostics-metrics-opentelemetry
- https://learn.microsoft.com/en-us/dotnet/core/diagnostics/metrics-collection
- https://learn.microsoft.com/en-us/dotnet/core/diagnostics/observability-with-otel
- https://learn.microsoft.com/en-us/dotnet/core/diagnostics/metrics-instrumentation
- https://www.mytechramblings.com/posts/getting-started-with-opentelemetry-metrics-and-dotnet-part-2/
- https://www.mytechramblings.com/posts/getting-started-with-opentelemetry-metrics-and-dotnet-part-1/
- https://github.com/prometheus-net/prometheus-net#summary

Worker saturation (ratio between request to the worker and pod uptime):
```prom
delta(TaskHandler_RequestExecution_milliseconds_sum[1m]) / delta(uptime_Milliseconds_total[1m])
```

Overlapping rate:
```
delta(TaskHandler_RequestExecution_milliseconds_sum[1m]) / (delta(TaskHandler_AcquireTask_milliseconds_sum[1m] + TaskHandler_PreProcessing_milliseconds_sum[1m] + TaskHandler_ExecuteTask_milliseconds_sum[1m] + TaskHandler_PostProcessing_milliseconds_sum[1m] + TaskHandler_DisposeAsync_milliseconds_sum[1m]))
```


Common/src/Meter/FunctionExecutionMetrics.cs
```diff
@@ -0,0 +1,70 @@
...
+using ArmoniK.Utils;
+
+namespace ArmoniK.Core.Common.Meter;
+
+public class FunctionExecutionMetrics
+{
+  private readonly string                                                                          className_;
+  private readonly ConcurrentDictionary<string, (Counter<int> counter, Histogram<long> histogram)> instruments_ = new();
+  private readonly System.Diagnostics.Metrics.Meter                                                meter_;
+  private readonly KeyValuePair<string, object?>[]                                                 tags_;
```
lemaitre-aneo, line 34:
```suggestion
  private readonly IDictionary<string, object?>                                                 tags_;
```

Common/src/Pollster/AgentIdentifier.cs
```diff
@@ -0,0 +1,28 @@
...
+using System.Net;
+
+using ArmoniK.Core.Common.Utils;
+
+namespace ArmoniK.Core.Common.Pollster;
+
+public static class AgentIdentifier
+{
+  public static readonly string OwnerPodId   = LocalIpFinder.LocalIpv4Address();
+  public static readonly string OwnerPodName = Dns.GetHostName();
```
lemaitre-aneo, line 27: What if those throw on the control plane?

terraform/modules/monitoring/prometheus/inputs.tf
```diff
@@ -11,4 +11,9 @@ variable "exposed_port" {
   type    = number
   default = 9090
 
-}
\ No newline at end of file
+}
+
+variable "polling_agent_names" {
+  type    = list(string)
```
lemaitre-aneo, line 17:
```suggestion
  type    = set(string)
```

Compute/PollingAgent/src/ArmoniK.Core.Compute.PollingAgent.csproj
```diff
@@ -32,6 +32,7 @@
     <PackageReference Include="MongoDB.Driver.Core.Extensions.OpenTelemetry" Version="1.0.0" />
     <PackageReference Include="OpenTelemetry.Extensions.Hosting" Version="1.7.0" />
     <PackageReference Include="OpenTelemetry.Exporter.OpenTelemetryProtocol" Version="1.7.0" />
+    <PackageReference Include="OpenTelemetry.Exporter.Prometheus.AspNetCore" Version="1.7.0-rc.1" />
```
lemaitre-aneo, line 35: Do you really need to put a rc here?

Common/src/Pollster/AgentIdentifier.cs
aneojgurhem: Why should they ?

Compute/PollingAgent/src/ArmoniK.Core.Compute.PollingAgent.csproj
aneojgurhem: We already use it

Common/src/Pollster/AgentIdentifier.cs
aneojgurhem: I created a non-static class

terraform/modules/compute_plane/outputs.tf
```diff
@@ -0,0 +1,3 @@
+output "polling_agent_name" {
+  value = local.polling_agent_name
```
lemaitre-aneo, line 2: This would make Terraform aware of the dependency between the resource and the output.
You could even remove your local value with that.

```suggestion
  value = docker_container.polling_agent.name
```

lemaitre-aneo (dismissed):

lemaitre-aneo (approved):

PR #686, opened 2024-04-03
fix: add update debian base image in Dockerfile

ngruelaneo:


aneojgurhem (changes requested):
First, do it on each Dockerfile on the project.
Second, do not add a new layer.
Third, that's useless and should already be done by the base image

ngruelaneo:
> First, do it on each Dockerfile on the project. Second, do not add a new layer. Third, that's useless and should already be done by the base image

Will it be ok for you in a branch? The problem is that the base image are behind the security update... https://github.com/dotnet/dotnet-docker/issues/5325

aneojgurhem:
We will do a new release when it will be integrated in their base image

PR #691, opened 2024-04-12
fix: abort results when cancelling tasks

aneojgurhem:


Common/src/gRPC/Services/GrpcTasksService.cs
```diff
@@ -288,6 +288,14 @@ public class GrpcTasksService : Task.TasksBase
                                                                      ? System.Threading.Tasks.Task.CompletedTask
                                                                      : httpClient_.GetAsync("http://" + ownerPodId + ":1080/stopcancelledtask")).ConfigureAwait(false))
                        .ConfigureAwait(false);
+
+      await ResultLifeCycleHelper.AbortTasksAndResults(taskTable_,
+                                                       resultTable_,
+                                                       request.TaskIds,
+                                                       "Client requested cancellation",
```
lemaitre-aneo, line 295: I would add the request.TaskIds in the message:
```suggestion
                                                       $"Client requested cancellation of tasks {string.Join(", ", request.TaskIds)}",
```

lemaitre-aneo (dismissed):

lemaitre-aneo (approved):

PR #694, opened 2024-04-17
feat: Metrics for task running time

lemaitre-aneo:


Common/src/Pollster/TaskHandler.cs
```diff
@@ -142,6 +143,8 @@ public sealed class TaskHandler : IAsyncDisposable
     registration2_ = workerConnectionCts_.Token.Register(() => logger_.LogWarning("Cancellation triggered, start to properly cancel task"));
   }
 
+  public DateTime StartedAt { get; }
```
aneojgurhem, line 146:
```suggestion
  public readonly DateTime StartedAt = DateTIme.UtcNow;
```

aneojgurhem (approved):

Common/src/Pollster/TaskHandler.cs
lemaitre-aneo: Readonly member messes up the formatting, so I stayed with a property, But I took the idea of inline initialization.

aneojgurhem (approved):

PR #695, opened 2024-04-18
fix: set readiness check to degraded when pollster does not process tasks

aneojgurhem:


Common/src/Pollster/Pollster.cs
```diff
@@ -219,6 +219,11 @@ public async Task<HealthCheckResult> Check(HealthCheckTag tag)
       healthCheckFailedResult_ = result;
     }
 
+    if (worstStatus == HealthStatus.Healthy && tag == HealthCheckTag.Readiness && taskProcessingDict_.IsEmpty)
```
lemaitre-aneo, line 222:
```suggestion
    if (tag == HealthCheckTag.Readiness && taskProcessingDict_.IsEmpty)
```

lemaitre-aneo (approved):

PR #705, opened 2024-06-10
fix: Acknowledge message upon retry

lemaitre-aneo:


Common/src/Pollster/TaskHandler.cs
```diff
@@ -300,6 +300,12 @@ public async Task<AcquisitionStatus> AcquireTask()
                                                            CancellationToken.None)
                                      .ConfigureAwait(false);
 
+          // Propagate retried status to TaskHandler
+          taskData_ = taskData_ with
+                      {
+                        Status = TaskStatus.Retried,
```
aneojgurhem, line 306: I think it should be cancelled here because status was cancelling. Except if `ResultLifeCycleHelper.AbortTasksAndResults` always retries

Common/src/Pollster/TaskHandler.cs
```diff
@@ -447,6 +470,12 @@ public async Task<AcquisitionStatus> AcquireTask()
                                                          CancellationToken.None)
                                    .ConfigureAwait(false);
 
+        // Propagate retried status to TaskHandler
+        taskData_ = taskData_ with
+                    {
+                      Status = TaskStatus.Retried,
```
aneojgurhem, line 476: same here

Common/src/Pollster/TaskHandler.cs
lemaitre-aneo: I have to check, but either it is not tested, or it is actually retried, because the test check that taskData_ from task handler matches the DB.

aneojgurhem (approved):

PR #720, opened 2024-07-31
chore: better logs for UpdateDefinition

aneojgurhem:


Common/src/Storage/UpdateDefinition.cs
```diff
@@ -59,8 +61,11 @@ public class UpdateDefinition<T>
   public UpdateDefinition<T> Set<TProperty>(Expression<Func<T, TProperty>> property,
                                             TProperty                      value)
   {
-    Setters.Add(new PropertySet(RemovePropertyType(property),
-                                value));
+    Setters.Add(new PropertySet
+                {
+                  Value    = value,
```
ngruelaneo, line 66: this change in the order of the attribute does not seems to be coherent with the code after. I don't know enough of csharp so perhaps it doesn't matter.

aneojgurhem: with this way to create objects (records in this case), it does not matter anymore

ngruelaneo (approved):

PR #724, opened 2024-08-05
refactor: exception manager

lemaitre-aneo:


aneojgurhem (approved):

Common/src/Pollster/RunningTaskProcessor.cs
```diff
@@ -53,18 +53,22 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
...
     {
       try
       {
-        while (postProcessingTaskQueue_.RemoveException(out var exception))
+        TaskHandler taskHandler;
+        try
         {
-          runningTaskQueue_.AddException(exception);
+          taskHandler = await runningTaskQueue_.ReadAsync(Timeout.InfiniteTimeSpan,
+                                                          exceptionManager_.LateCancellationToken)
```
lemaitre-aneo, line 64: I think it should be the EarlyCancellationToken here

aneojgurhem (approved):

PR #738, opened 2024-08-27
fix: tasks will not be inserted in the queue during paused sessions

aneojgurhem:


Common/src/Storage/TaskLifeCycleHelper.cs
```diff
@@ -431,6 +432,7 @@ await taskTable.RemoveRemainingDataDependenciesAsync(dependentTasks,
 
     await EnqueueReadyTasks(taskTable,
                             pushQueueStorage,
+                            sessionData,
                             readyTasks,
                             cancellationToken)
       .ConfigureAwait(false);
```
lemaitre-aneo, line 438: Instead of checking if the session is paused within this function, I would check before calling the function.

Common/src/gRPC/Services/GrpcResultsService.cs
```diff
@@ -316,6 +321,8 @@ await objectStorage_.AddOrUpdateAsync(id.ResultId,
                                           context.CancellationToken)
                         .ConfigureAwait(false);
 
+    var sessionData = await sessionTask.ConfigureAwait(false);
```
lemaitre-aneo, line 324: What if this fails?

Common/tests/Helpers/TestPollsterProvider.cs
```diff
@@ -68,7 +68,7 @@ public class TestPollsterProvider : IDisposable
   public readonly  Common.Pollster.Pollster Pollster;
   public readonly  IResultTable             ResultTable;
   private readonly IMongoRunner             runner_;
-  private readonly ISessionTable            sessionTable_;
+  public readonly  ISessionTable            sessionTable_;
```
lemaitre-aneo, line 71:
```suggestion
  public readonly  ISessionTable            SessionTable;
```

Common/tests/Pollster/AgentTest.cs
```diff
@@ -136,6 +136,7 @@ private class AgentHolder : IDisposable
     public readonly  MyPushQueueStorage   QueueStorage;
     public readonly  IResultTable         ResultTable;
     public readonly  string               Session;
+    private readonly SessionData          sessionData;
```
lemaitre-aneo, line 139:
```suggestion
    public readonly SessionData          SessionData;
```

Common/src/Storage/TaskLifeCycleHelper.cs
aneojgurhem: The function do other things that we want to keep doing so we need to pass the info to the function

PR #740, opened 2024-08-28
fix: set result associated to payload to deleted when the payload is deleted

aneojgurhem:


Common/src/Storage/ResultTableExtensions.cs
```diff
@@ -255,4 +255,22 @@ public static async Task<IEnumerable<ResultIdStatus>> GetResultStatus(this IResu
...
+  /// <summary>
+  ///   Set the result status to <see cref="ResultStatus.DeletedData" />
+  /// </summary>
+  /// <param name="resultTable">Interface to manage results</param>
+  /// <param name="resultId">Id of the result to update</param>
+  /// <param name="cancellationToken">Token used to cancel the execution of the method</param>
+  /// <returns>
+  ///   Task representing the asynchronous execution of the method
+  /// </returns>
+  public static async Task MarkAsDeleted(this IResultTable resultTable,
```
lemaitre-aneo, line 268: I am not sure it is needed to create an extension method for that as it is used only once in the code.

lemaitre-aneo (approved):

Common/src/Storage/ResultTableExtensions.cs
aneojgurhem: That was done like that for the others

PR #746, opened 2024-09-02
feat: add Pending and Paused status

aneojgurhem:


Common/src/Storage/TaskLifeCycleHelper.cs
```diff
@@ -246,10 +246,19 @@ await resultTable.ChangeResultOwnership(parentTaskId,
...
     }
 
+    var readyTask = await prepareTaskDependencies.ConfigureAwait(false);
+    await taskRequests.Select(request => request.TaskId)
+                      .Chunk(100)
+                      .ParallelForEach(strings => taskTable.UpdateManyTasks(data => data.Status == TaskStatus.Creating && strings.Contains(data.TaskId),
+                                                                            new UpdateDefinition<TaskData>().Set(data => data.Status,
+                                                                                                                 TaskStatus.Pending),
+                                                                            cancellationToken))
+                      .ConfigureAwait(false);
```
lemaitre-aneo, line 256: We did not chunk those kind of requests in the rest of the code. I am not sure of the impact, though.

Common/src/Storage/TaskLifeCycleHelper.cs
```diff
@@ -424,7 +433,7 @@ await taskTable.RemoveRemainingDataDependenciesAsync(dependentTasks,
     // Find all tasks whose dependencies are now complete in order to start them.
     // Multiple agents can see the same task as ready and will try to start it multiple times.
     // This is benign as it will be handled during dequeue with message deduplication.
-    var readyTasks = await taskTable.FindTasksAsync(data => dependentTasks.Contains(data.TaskId) && data.Status == TaskStatus.Creating &&
+    var readyTasks = await taskTable.FindTasksAsync(data => dependentTasks.Contains(data.TaskId) && data.Status == TaskStatus.Pending &&
```
lemaitre-aneo, line 436: I wonder if you should keep the test on Creating as well.

Common/src/Pollster/TaskHandler.cs
```diff
@@ -652,21 +660,24 @@ public TaskInfo GetAcquiredTaskInfo()
   ///   Task representing the asynchronous execution of the method
   /// </returns>
   /// <exception cref="NullReferenceException">wrong order of execution</exception>
-  public async Task ReleaseAndPostponeTask()
+  public async Task ReleaseAndPostponeTask(bool paused = false)
```
lemaitre-aneo, line 663: xmldoc

Common/src/Pollster/TaskHandler.cs
```diff
@@ -652,21 +660,24 @@ public TaskInfo GetAcquiredTaskInfo()
...
     // Propagate submitted status to TaskHandler
     taskData_ = taskData_ with
                 {
-                  Status = TaskStatus.Submitted,
+                  Status = paused
+                             ? TaskStatus.Paused
+                             : TaskStatus.Submitted,
                 };
 
     messageHandler_.Status = QueueMessageStatus.Postponed;
```
lemaitre-aneo, line 683: If it is paused, the message should not be postponed, but acknowledged

lemaitre-aneo (commented):
I suspect there could be a race condition between task creation / result completion, and pause of a session. But I don't have time to check for now.

Common/src/Exceptions/TaskPausedException.cs
```diff
@@ -0,0 +1,40 @@
...
+  {
+  }
+
+  public TaskPausedException(string    message,
+                             Exception innerException)
+    : base(message,
+           innerException)
+  {
+  }
+}
```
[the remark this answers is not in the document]
aneojgurhem: If this work, I'll also do it for the other in another pr

Common/src/Storage/TaskLifeCycleHelper.cs
aneojgurhem: I'll remove the chunk for consistency purpose. To test further

Common/src/Storage/TaskLifeCycleHelper.cs
aneojgurhem: We do not really need it. Except for two things:
- if old messages are in the queue (should not happen much)
- if we want to save a request that changes the task status to Pending when we know it will be changed to Submitted shortly after

Common/src/Storage/TaskLifeCycleHelper.cs
```diff
@@ -476,10 +485,22 @@ await pushQueueStorage.PushMessagesAsync(group,
...
+  ///   Resume session and its paused tasks
+  /// </summary>
+  /// <param name="taskTable">Interface to manage task states</param>
+  /// <param name="sessionTable">Interface to manage session states</param>
+  /// <param name="pushQueueStorage">Interface to push tasks in the queue</param>
+  /// <param name="sessionId">Id of the session to resume</param>
+  /// <param name="cancellationToken">Token used to cancel the execution of the method</param>
+  /// <returns>
+  ///   The updated data of the session
+  /// </returns>
```
dbrasseur-aneo, line 503: The tasks are pushed to the queue and then their status is updated. A race condition could occur if some polling agent pulls a task from the queue before the said task is switched to "submitted". The polling agent would then remove it off the queue without putting it back

Common/src/Exceptions/TaskPausedException.cs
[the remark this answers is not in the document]
aneojgurhem: It did not work

aneojgurhem:
> I suspect there could be a race condition between task creation / result completion, and pause of a session. But I don't have time to check for now.

I think we keep a consistent state of the session. We read it only once so if it changes during we keep the old state. Marginally, there will be some tasks submitted when the session is paused and some tasks will not be notified the session has been resumed.

.github/workflows/build.yml
```diff
@@ -791,6 +791,16 @@ jobs:
             -e BenchOptions__TaskDurationMs=10 \
             -e GrpcClient__Endpoint=http://armonik.control.submitter:1080 \
             dockerhubaneo/armonik_core_bench_test_client:$VERSION
+      - name: Run Bench test tasks - many tasks (400) wuth pause
```
ngruelaneo, line 794:
```suggestion
      - name: Run Bench test tasks - many tasks (400) with pause
```

Common/src/Storage/TaskLifeCycleHelper.cs
```diff
@@ -499,13 +521,51 @@ public static async Task<SessionData> ResumeAsync(ITaskTable        taskTable,
...
+                                .ConfigureAwait(false);
+
+      var taskIds = tasks.Select(task => task.TaskId)
+                         .AsICollection();
+      await taskTable.UpdateManyTasks(data => data.SessionId == sessionId && data.Status == TaskStatus.Paused && taskIds.Contains(data.TaskId),
+                                      new UpdateDefinition<TaskData>().Set(data => data.Status,
+                                                                           TaskStatus.Submitted),
+                                      cancellationToken)
+                     .ConfigureAwait(false);
+      await pushQueueStorage.PushMessagesAsync(tasks,
```
dbrasseur-aneo, line 534: If the control plane hard crashes between these, then it can be fixed by pausing then unpausing the session

dbrasseur-aneo (approved):

PR #762, opened 2024-09-24
fix: resubmit task when agent tries to acquire a task in Retried

aneojgurhem:
# Motivation

Fix the issue in which retried task appear in Creating and are not processed further, preventing application from running to completion.

To my understanding, during downscaling, preemption or crash and when the worker crashes too, the agent can be abruptly stopped during the submission of the retried task (task creation, task finalization or insertion into the queue). We also observe that the agent tries to acquire the initial task with status Retried instead of the new copy of the task (the new attempt). The retried task is only in the Creating status, showing that the task finalization was not properly done. This means that the initial task' finalization has not completed properly.

# Description

To mitigate this issue, we implement a recovery mechanism that, when an agent tries to acquire the initial task, the agent will properly finalize the retried task before removing the initial task from the queue instead of only removing the initial task from the queue.

When the initial task with status Retried is acquired by the agent, we check whether the retry task has been properly finalized. To do so, we get the metadata of the retried task. If we can retrieve them and the retried task is in Creating or Submitted we perform task finalization and insertion in the queue again. If the task was already inserted in the queue, task deduplication should do its job and ignore the duplicate. If the retried task is in another status, we remove the message from the queue. If the retried task is not found, we submit the retried task completely (creation, finalization, queueing). If the retried task has already been created between our read and our creation in the database, we check the status of the retried tasks and perform finalization if task is Creating or Submitted.

# Testing

I was not able to fully reproduce this issue in Core docker deployment even while trying to stop agents abruptly with the following script. I used a modified bench worker that was only producing errors.

```bash
#!/bin/sh

for i in $(seq 1 40); do
    for a in $(docker ps -q --filter name=armonik.compute.pollingagent); do
        docker restart -s sigterm -t 0 $a
        # sleep 1
    done
done
```

I also tried to vary the delay and still was not able to reproduce.

I added unit tests that can put tasks in the same state that we observed. They also validate that we are able to recover from the invalid state and resubmit the task in retry.

# Impact

- Acquisition of task in Retried status is more complex now and make more calls to the database, reducing performances while improving recovery on failure.
- Sometimes, in the case explicited here, the retry task may be inserted twice into the queue, making use of the deduplication mechanism to remove the duplicata from the queue.

# Checklist

### DIFF
--- a/Adaptors/MongoDB/src/TaskTable.cs
+++ b/Adaptors/MongoDB/src/TaskTable.cs
@@ -81,7 +81,7 @@ public class TaskTable : ITaskTable
                                            cancellationToken: cancellationToken)
                           .ConfigureAwait(false);
     }
-    catch (MongoBulkWriteException<TaskData> e)
+    catch (MongoBulkWriteException<TaskData> e) when (e.WriteErrors.All(error => error.Category == ServerErrorCategory.DuplicateKey))
     {
       throw new TaskAlreadyExistsException("Task already exists",
                                            e);
--- a/Adaptors/MongoDB/src/TaskTable.cs
+++ b/Adaptors/MongoDB/src/TaskTable.cs
@@ -75,9 +75,17 @@ public class TaskTable : ITaskTable
 
     var taskCollection = taskCollectionProvider_.Get();
 
-    await taskCollection.InsertManyAsync(tasks.Select(taskData => taskData),
-                                         cancellationToken: cancellationToken)
-                        .ConfigureAwait(false);
+    try
+    {
+      await taskCollection.InsertManyAsync(tasks.Select(taskData => taskData),
+                                           cancellationToken: cancellationToken)
+                          .ConfigureAwait(false);
+    }
+    catch (MongoBulkWriteException<TaskData> e)
+    {
+      throw new TaskAlreadyExistsException("Task already exists",
+                                           e);
+    }
   }
 
   /// <inheritdoc />
--- a/Common/src/Exceptions/TaskAlreadyExistsException.cs
+++ b/Common/src/Exceptions/TaskAlreadyExistsException.cs
@@ -1,0 +1,40 @@
+// This file is part of the ArmoniK project
+// 
+// Copyright (C) ANEO, 2021-2024. All rights reserved.
+// 
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as published
+// by the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+// 
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY, without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+// 
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+using System;
+
+namespace ArmoniK.Core.Common.Exceptions;
+
+[Serializable]
+public class TaskAlreadyExistsException : ArmoniKException
+{
+  public TaskAlreadyExistsException()
+  {
+  }
+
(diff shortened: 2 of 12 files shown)


Adaptors/MongoDB/src/TaskTable.cs
```diff
@@ -75,9 +75,17 @@ public async Task CreateTasks(IEnumerable<TaskData> tasks,
...
-    await taskCollection.InsertManyAsync(tasks.Select(taskData => taskData),
-                                         cancellationToken: cancellationToken)
-                        .ConfigureAwait(false);
+    try
+    {
+      await taskCollection.InsertManyAsync(tasks.Select(taskData => taskData),
+                                           cancellationToken: cancellationToken)
+                          .ConfigureAwait(false);
+    }
+    catch (MongoBulkWriteException<TaskData> e)
```
lemaitre-aneo, line 84: Isn't this catch too wide?

Common/src/Exceptions/TaskAlreadyExistsException.cs
```diff
@@ -0,0 +1,40 @@
...
+  {
+  }
+
+  public TaskAlreadyExistsException(string    message,
+                                    Exception innerException)
+    : base(message,
+           innerException)
+  {
+  }
+}
```
lemaitre-aneo, line 40:
```suggestion
public class TaskAlreadyExistsException : ArmoniKException;
```

Common/src/Pollster/TaskHandler.cs
```diff
@@ -445,6 +446,79 @@ await submitter_.CompleteTaskAsync(taskData_,
...
+                                                      },
+                                                      sessionData_,
+                                                      taskData_.TaskId,
+                                                      CancellationToken.None)
+                                .ConfigureAwait(false);
+              }
+            }
+
+            return AcquisitionStatus.TaskIsRetriedAndRetryIsNotFound;
+          }
```
lemaitre-aneo, line 520: I think you can factorize a bit more this code using the following pattern:

```c#
TaskData retryData;
try {
  retryData = await taskTable_.ReadTaskAsync(...);
} catch (TaskNotFoundException) {
  logger_.LogWarning("Retried task {task} was not found in the database; resubmit it", ...);
  try {
    taskTable_.RetryTaskAsync(...);
  } catch (TaskAlreadyException) {
    logger_.LogWarning("Retried task {task} already exists; finalize creation if needed", ...);
  }
  retryData = await taskTable_.ReadTaskAsync(...);
}
if (retryData.Status is TaskStatus.Creating or TaskStatus.Submitted)
{
  logger_.LogWarning("Retried task {task} is in {status}; trying to finalize task creation", ...);
  await submitter_.FinalizeTaskCreation(...);
}
```

Common/src/Exceptions/TaskAlreadyExistsException.cs
aneojgurhem: Does not compile

Common/src/Pollster/TaskHandler.cs
```diff
@@ -490,36 +472,47 @@ await taskTable_.RetryTask(taskData_,
...
+                                                        retryData.ExpectedOutputIds,
+                                                        retryData.DataDependencies),
+                                                  },
+                                                  sessionData_,
+                                                  taskData_.TaskId,
+                                                  CancellationToken.None)
+                            .ConfigureAwait(false);
+          }
+
+          switch ((taskNotFound, taskExists, retryData.Status))
```
lemaitre-aneo, line 502: I think switch expression would be simpler, and I think you can select on multiple patterns with `or`

Common/src/Exceptions/TaskAlreadyExistsException.cs
lemaitre-aneo: weird. Maybe t's from a new C# version. But `{}` should work.

aneojgurhem: Missing constructors now

lemaitre-aneo (approved):

PR #789, opened 2024-10-30
fix: cancelling session from sessions service cancel tasks and abort results from the session

aneojgurhem:
# Motivation

Previously, cancelling a session from sessions service was only updating the session status, leaving all its submitted and running tasks in autoscaling count. This PR removes tasks of a cancelled session from the autoscaling count by cancelling of the tasks and results from the session.

# Description

When calling the Cancel RPC from the session service, we also change task and results statuses in the cancelled session.

# Testing

On an ArmoniK instance, this new implementation was used and confirmed to change the statuses.

# Impact

- It will reduce the computing resources consummed because the submitted tasks from a cancelled session will change status to cancelling and not count for autoscaling.
- If previously cancelling tasks are not removed from the queue during downscaling, when a new session is submitted and upscaling occurs, cancelling tasks will be removed first from the queue storage before new tasks will be processed (at equal priority).

# Checklist

### DIFF
--- a/Common/src/gRPC/Services/GrpcSessionsService.cs
+++ b/Common/src/gRPC/Services/GrpcSessionsService.cs
@@ -81,11 +81,18 @@ public class GrpcSessionsService : Sessions.SessionsBase
     using var measure = meter_.CountAndTime();
     try
     {
+      var tasks = taskTable_.CancelSessionAsync(request.SessionId,
+                                                context.CancellationToken);
+      var results = resultTable_.AbortSessionResults(request.SessionId,
+                                                     context.CancellationToken);
+      var sessions = sessionTable_.CancelSessionAsync(request.SessionId,
+                                                      context.CancellationToken);
+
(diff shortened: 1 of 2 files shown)


Common/src/gRPC/Services/GrpcSessionsService.cs
```diff
@@ -81,6 +81,14 @@ public override async Task<CancelSessionResponse> CancelSession(CancelSessionReq
...
     try
     {
+      await taskTable_.CancelSessionAsync(request.SessionId,
+                                          context.CancellationToken)
+                      .ConfigureAwait(false);
+
+      await resultTable_.AbortSessionResults(request.SessionId,
+                                             context.CancellationToken)
+                        .ConfigureAwait(false);
+
```
lemaitre-aneo, line 91: I would start all three tasks in parallel

lemaitre-aneo (dismissed):

ngruelaneo (approved):

PR #799, opened 2024-11-18
feat: Added a python script utility to import logs stored in an S3 bucket to Seq 

AncientPatata:
# Motivation

This PR is related to the work done to [export logs from localhost to an S3 bucket](https://github.com/aneoconsulting/ArmoniK.Infra/pull/187). Adding a more general script to import saved ArmoniK logs from previous runs to a Seq deployment for further analysis. 

# Description

Building on the log2seq.py script from #62, generalized it to download entire directories that you can freely specify to match the new exports.

# Testing

Multiple test runs of the script.

# Impact

Not applicable.

# Additional Information

[Any additional information that reviewers should be aware of.]

# Checklist

### DIFF
--- a/tools/s3_to_seq/s3_to_seq.py
+++ b/tools/s3_to_seq/s3_to_seq.py
@@ -1,0 +1,326 @@
+import argparse
+import gzip
+import json
+import logging
+import shutil
+from pathlib import Path
+from typing import List
+
+import boto3
+import botocore
+import botocore.exceptions
+import filetype
+import requests
+from colorama import Back, Fore, Style, init
+
+
+init(autoreset=True)
+
+
+class ColoredFormatter(logging.Formatter):
+    """
+    A custom logging formatter that adds color to log level names based on severity.
+
+    Attributes:
+        COLORS (dict): Maps log levels to ANSI color codes for formatted output.
+    """
+
+    COLORS = {
+        "DEBUG": Fore.CYAN,
+        "INFO": Fore.GREEN,
+        "WARNING": Fore.YELLOW,
+        "ERROR": Fore.RED,
+        "CRITICAL": Fore.RED + Back.WHITE + Style.BRIGHT,
+    }
+
+    def format(self, record):
+        levelname = record.levelname
+        if levelname in self.COLORS:
+            levelname_color = self.COLORS[levelname] + levelname + Style.RESET_ALL
+            record.levelname = levelname_color
+        return super().format(record)
+
+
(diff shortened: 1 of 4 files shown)


tools/s3_to_seq/s3_to_seq.py
```diff
@@ -0,0 +1,328 @@
...
+                ctx.logger.error(
+                    f"Attempt {attempt} at downloading file {obj.key} failed with error: \n {json.dumps(e.response['Error'])} "
+                )
+            ctx.logger.warning(f"All attempts at downloading {obj.key} failed, skipping..")
+    return file_list
+
+
+class LogSender:
+    def __init__(self, url: str):
+        """
```
qdelamea-aneo, line 162: Move the docstring after the class definition.

tools/s3_to_seq/s3_to_seq.py
```diff
@@ -0,0 +1,328 @@
...
+    """
+    ctx = Context()
+    ctx.logger.debug("Sending gzipped json file : " + file_name)
+    with gzip.open(file_name, "r") as file:
+        with LogSender(url) as log_sender:
+            for line in file.read().decode("utf-8").split("\n"):
+                log_sender.sendlog(line)
+
+
+def main(args: argparse.Namespace):
```
qdelamea-aneo, line 252: Do not use argument namespace as function input. Instead list arguments and then call the function using **args to unpack the namespace.

tools/s3_to_seq/s3_to_seq.py
```diff
@@ -0,0 +1,328 @@
...
+    logger = LoggingWrapper(args.log_level)
+    logger.info("S3 to Seq started")
+    context = Context(logger, Config(args.temp_path, args.nattempts))
+
+    logger.info("Temporary download directory set to " + args.temp_path)
+
+    try:
+        try:
+            s3 = boto3.client("s3")
+        except Exception:
```
qdelamea-aneo, line 262: When handling exceptions you must always catch precise exceptions. Here you catch any exception even if it its not related to the client creation (like SIGTERM). So replace with the precise exeption(s) you wan to catch.

qdelamea-aneo (changes requested):

aneojgurhem (commented):
code is great but I would have expected to see this in ArmoniK instead of ArmoniK.Core

AncientPatata:
> code is great but I would have expected to see this in ArmoniK instead of ArmoniK.Core

Moving it

PR #818, opened 2025-01-14
feat: allow usage of tcp socket between agent and worker

jfonseca-aneo:
# Motivation

Java does not support grpc communication between the worker and the agent over unix sockets, this PR builds on the changes introduced in https://github.com/aneoconsulting/ArmoniK.Api/pull/583 to enable the option to use a tcp socket instead.

# Description

Mimic the changes in https://github.com/aneoconsulting/ArmoniK.Api/pull/583 and enlarged the kestrel configuration options in the `AgentHandler` such that it also accepts to listen on a tcp socket. In addition this PR separates the volumes used for shared data and unix socket. Before the PR both of them where mounted on `\cache`,  the updated mount point for shared data is `\comm`.   

# Testing

Terraform deployment and `just` file have been modified to take into account a new variable `socket_type`,  that can be set to either `unixdomainsocket` or `tcp` so the deployment configures which socket type should be employed accordingly. With these changes the same set of tests that ran for unix sockets can be executed but using a tcp socket instead . 

# Impact

The desired life cycle behavior in a deployment with the whole infrastructure is that the worker starts before the agent and stops after it. This is currently enforced by checking the absence/existence of the unix socket created by the agent. With the changes of this PR, in the case of choosing a tcp socket for the connection, the unix socket is not created. Hence, the infra should be adapted accordingly to guarantee the desired worker/agent life cycle.  One option is to transform the worker container in Kubernetes POD into a sidecar container, but this will require the direct use of Helm charts since the Kubernetes provider we use for terraform does not support this yet.

# Additional Information

Not applicable

# Checklist

### DIFF
--- a/terraform/variables.tf
+++ b/terraform/variables.tf
@@ -126,6 +126,7 @@ variable "compute_plane" {
       // They will be used for both
       shared_socket = optional(string, "/cache")
       shared_data   = optional(string, "/cache")
+      socket_type   = optional(string, "unixdomainsocket")
     })
   })
   default = {
--- a/terraform/variables.tf
+++ b/terraform/variables.tf
@@ -106,6 +106,16 @@ variable "worker_image" {
   default = "dockerhubaneo/armonik_core_htcmock_test_worker"
 }
 
+variable "socket_type" {
+  type        = string
+  description = "Socket type used by agent and worker to communicate"
+  validation {
+    condition     = can(regex("^(unixdomainsocket|tcp)$", var.socket_type))
+    error_message = "Socket must be either unixdomainsocket or tcp"
+  }
+  default = "unixdomaisocket"
+}
+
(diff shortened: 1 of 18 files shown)


terraform/variables.tf
```diff
@@ -106,6 +106,16 @@ variable "worker_image" {
...
 }
 
+variable "socket_type" {
+  type        = string
+  description = "Socket type used by agent and worker to communicate"
+  validation {
+    condition     = can(regex("^(unixdomainsocket|tcp)$", var.socket_type))
+    error_message = "Socket must be either unixdomainsocket or tcp"
+  }
+  default = "unixdomaisocket"
```
ngruelaneo, line 116:
```suggestion
  default = "unixdomainsocket"
```

aneojgurhem (approved):

PR #831, opened 2025-01-29
fix: unable to acquire task when a release happens during its re-acquisition

italo1aguiar-aneo:
# Motivation

Being able to acquire tasks even though it is released while the re-acquisition.

# Description

When a task is acquired it might be a task from a pod that crashed. During this acquisition it is possible that the task status will change as the pod releases the task. This scenario is not being treated by the actual code.

# Testing

[When applicable, detail the testing you have performed to ensure that these changes function as intended. Include information about any added tests.]

To do

# Impact

It fixes the behaviour of task that are not processed. 

# Checklist


Common/src/Pollster/AcquisitionStatus.cs
```diff
@@ -177,4 +177,9 @@ public enum AcquisitionStatus
   ///   Task is missing dependencies but happens to be in the queue
   /// </summary>
   TaskIsPending,
+
+  /// <summary>
+  ///   Task is submitted but execution is no possible
```
ngruelaneo, line 182:
```suggestion
  ///   Task is submitted but execution is not possible
```

Common/src/Pollster/TaskHandler.cs
```diff
@@ -626,8 +626,14 @@ await ReleaseAndPostponeTask()
           logger_.LogInformation("Task is not running on the other polling agent, status : {status}",
                                  taskData_.Status);
 
-          if (taskData_.Status is TaskStatus.Dispatched && taskData_.AcquisitionDate + delayBeforeAcquisition_ > DateTime.UtcNow)
+          if (taskData_.Status is TaskStatus.Submitted)
+          {
+            messageHandler_.Status = QueueMessageStatus.Postponed;
+            // TODO: AcquistionStaus must be tested
```
ngruelaneo, line 632:
```suggestion
            // TODO: AcquistionStatus must be tested
```

aneojgurhem (approved):

PR #838, opened 2025-02-05
fix: add validation for certificate name mismatch in TLS validation

Nico-dl05:
# Motivation

We added a check for **RemoteCertificateNameMismatch** to further secure our TLS validation and avoid potential mismatches between the certificate name and the domain in use. We also introduced a boolean **allowHostMismatch** that lets users choose between strict domain name matching or allowing a mismatch between the certificate and the domain.

# Description

We updated the SSL validation logic to handle RemoteCertificateNameMismatch conditionally.

   - When allowHostMismatch is false, we reject certificates that present a mismatched domain name.
   - When allowHostMismatch is true, we tolerate this error and remove it from the list of SSL policy errors.

# Testing

- Verified that certificates with mismatched names are rejected if allowHostMismatch = false.
- Ensured mismatched certificates are allowed if allowHostMismatch = true.

# Impact

- Allows flexible TLS configuration based on different environment needs.

# Additional Information

None

# Checklist

### DIFF
--- a/Adaptors/Redis/src/Options/Redis.cs
+++ b/Adaptors/Redis/src/Options/Redis.cs
@@ -22,17 +22,18 @@ namespace ArmoniK.Core.Adapters.Redis.Options;
 public class Redis
 {
   public const string   SettingSection = nameof(Redis);
-  public       string   InstanceName    { get; set; } = "";
-  public       string   EndpointUrl     { get; set; } = "";
-  public       string   ClientName      { get; set; } = "";
-  public       string   SslHost         { get; set; } = "";
-  public       int      Timeout         { get; set; }
-  public       string   Password        { get; set; } = "";
-  public       string   User            { get; set; } = "";
-  public       bool     Ssl             { get; set; }
-  public       string   CredentialsPath { get; set; } = "";
-  public       string   CaPath          { get; set; } = "";
-  public       int      MaxRetry        { get; set; } = 5;
-  public       int      MsAfterRetry    { get; set; } = 500;
-  public       TimeSpan TtlTimeSpan     { set; get; } = TimeSpan.MaxValue;
+  public       string   InstanceName      { get; set; } = "";
+  public       string   EndpointUrl       { get; set; } = "";
+  public       string   ClientName        { get; set; } = "";
+  public       string   SslHost           { get; set; } = "";
+  public       int      Timeout           { get; set; }
+  public       string   Password          { get; set; } = "";
+  public       string   User              { get; set; } = "";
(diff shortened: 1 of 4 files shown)


Adaptors/Redis/src/Options/Redis.cs
```diff
@@ -22,17 +22,18 @@ namespace ArmoniK.Core.Adapters.Redis.Options;
...
+  public       int      Timeout           { get; set; }
+  public       string   Password          { get; set; } = "";
+  public       string   User              { get; set; } = "";
+  public       bool     Ssl               { get; set; }
+  public       string   CredentialsPath   { get; set; } = "";
+  public       string   CaPath            { get; set; } = "";
+  public       int      MaxRetry          { get; set; } = 5;
+  public       int      MsAfterRetry      { get; set; } = 500;
+  public       TimeSpan TtlTimeSpan       { set; get; } = TimeSpan.MaxValue;
+  public       bool     allowHostMismatch { get; set; }
```
aneojgurhem, line 38:
```suggestion
  public       bool     AllowHostMismatch { get; set; }
```

terraform/modules/storage/object/redis/outputs.tf
```diff
@@ -12,6 +12,7 @@ output "generated_env_vars" {
     "Components__ObjectStorageAdaptorSettings__ClassName"           = "ArmoniK.Core.Adapters.Redis.ObjectBuilder"
     "Components__ObjectStorageAdaptorSettings__AdapterAbsolutePath" = "/adapters/object/redis/ArmoniK.Core.Adapters.Redis.dll"
     "Redis__EndpointUrl"                                            = "redis:6380"
+    "Redis__AllowInsecureTls"                                       = "false"
```
aneojgurhem, line 15: This is the default value, thus you can remove it

Utils/src/ServerCertificateValidator.cs
```diff
@@ -51,10 +52,32 @@ public static RemoteCertificateValidationCallback ValidationCallback(ILogger
...
+                               sslPolicyErrors);
+             return false;
+           }
+
+           // We allow host mismatch, so we remove the error
+           sslPolicyErrors &= ~SslPolicyErrors.RemoteCertificateNameMismatch;
+           logger.LogDebug("Allowing RemoteCertificateNameMismatch since allowHostMismatch=true");
+         }
+
+         // If the certificate name does not match the server name, fail the validation
```
aneojgurhem, line 70: this comment seems wrong

aneojgurhem (approved):

PR #866, opened 2025-02-26
refactor: create child tasks and results during post processing instead of during task execution

aneojgurhem:
# Motivation

Previously, child tasks and results were generated during the execution of a parent task, which could lead to complexities such as having to cancel child tasks when the parent task fails. By shifting this creation to the post-processing stage, we aim to streamline task management and improve overall system stability.

# Description

This PR refactors the task processing workflow by moving the creation of child tasks and their corresponding results from the task execution phase to the post-processing phase. This change aims to enhance the efficiency and clarity of the task management process.

# Testing

- Unit tests were updated and executed to ensure all components function as expected.
- Integration tests were performed to validate the end-to-end workflow with the new post-processing logic.

# Impact

- **Performance Improvement**: By moving child task and result creation to post-processing, task execution becomes more efficient, reducing potential bottlenecks.
- **Better Task Management**: This change clarifies the task lifecycle, making it easier to track dependencies and execution order.

# Checklist


Common/src/Pollster/Agent.cs
```diff
@@ -47,16 +48,18 @@ namespace ArmoniK.Core.Common.Pollster;
...
-  private readonly IObjectStorage                             objectStorage_;
-  private readonly IPushQueueStorage                          pushQueueStorage_;
-  private readonly IResultTable                               resultTable_;
-  private readonly Dictionary<string, (byte[] id, long size)> sentResults_;
-  private readonly SessionData                                sessionData_;
-  private readonly ISubmitter                                 submitter_;
-  private readonly TaskData                                   taskData_;
-  private readonly ITaskTable                                 taskTable_;
+  private readonly ConcurrentBag<Result>                              createdResults_;
+  private readonly ConcurrentBag<TaskCreationRequest>                 createdTasks_;
```
lemaitre-aneo, line 52: I think those could be `ConcurrentBag<List<?>>` and flattened in the post processing to reduce the cost of adding multiple elements inside the bag and iterating over it.

Common/src/Pollster/Agent.cs
```diff
@@ -116,27 +121,90 @@ public async Task FinalizeTaskCreation(CancellationToken cancellationToken)
 
     logger_.LogDebug("Finalize child task creation");
```
lemaitre-aneo, line 122: I think the name of the method is wrong now, as this method implements much more than just finalizing task creation. Maybe it could be split into multiple methods. I think you could call directly the submitter steps instead of reimplementing them.

Common/src/Pollster/Agent.cs
```diff
@@ -116,27 +121,90 @@ public async Task FinalizeTaskCreation(CancellationToken cancellationToken)
...
-                                        result,
-                                        size,
-                                        id,
-                                        cancellationToken)
-                        .ConfigureAwait(false);
+      await using var fs = new FileStream(Path.Combine(Folder,
+                                                       result),
+                                          FileMode.OpenOrCreate);
+      using var r       = new BinaryReader(fs);
+      var       channel = Channel.CreateUnbounded<ReadOnlyMemory<byte>>();
```
lemaitre-aneo, line 151: I don't think that's a good idea to have an unbounded channel here. We don't want to materialize the whole file in memory in case the upload to the Object Storage is slower than reading the file from disk (which it most likely is). I don't know what the capacity should be, but I think a handful of chunk is enough.

```suggestion
      var       channel = Channel.CreateBounded<ReadOnlyMemory<byte>>(5);
```

Common/src/Pollster/Agent.cs
```diff
@@ -116,27 +121,90 @@ public async Task FinalizeTaskCreation(CancellationToken cancellationToken)
...
     await submitter_.FinalizeTaskCreation(createdTasks_,
                                           sessionData_,
                                           taskData_.TaskId,
                                           cancellationToken)
                     .ConfigureAwait(false);
 
-    foreach (var (result, (id, size)) in sentResults_)
+    var resultsToComplete = new Dictionary<string, (byte[] id, long size)>();
+
+    foreach (var result in notifiedResults_)
```
lemaitre-aneo, line 145: This could most likely be its own method.

Common/src/Pollster/Agent.cs
```diff
@@ -116,27 +121,90 @@ public async Task FinalizeTaskCreation(CancellationToken cancellationToken)
...
+        }
+      } while (read != 0);
+
+      channel.Writer.Complete();
+
+      var add = await addTask.ConfigureAwait(false);
+      resultsToComplete[result] = add;
+    }
+
+    foreach (var (resultId, memory) in resultsData_)
```
lemaitre-aneo, line 183: This could also be its own method.

Common/src/Pollster/Agent.cs
```diff
@@ -233,7 +301,7 @@ public async Task<ICollection<TaskCreationRequest>> SubmitTasks(ICollection<Task
 
     if (requests.Count == 0)
     {
-      return new List<TaskCreationRequest>();
+      return Task.FromResult<ICollection<TaskCreationRequest>>(new List<TaskCreationRequest>());
```
lemaitre-aneo, line 304:
```suggestion
      return Task.FromResult(Array.Empty<TaskCreationRequest>().AsICollection());
```

Common/src/Pollster/Agent.cs
aneojgurhem: I agree, the name does not show what is done. I will change it.

I already use what is available. I even already use Submitter itself.

Common/src/Pollster/Agent.cs
```diff
@@ -47,16 +48,18 @@ namespace ArmoniK.Core.Common.Pollster;
...
-  private readonly ISubmitter                                 submitter_;
-  private readonly TaskData                                   taskData_;
-  private readonly ITaskTable                                 taskTable_;
+  private readonly ConcurrentBag<ICollection<Result>>               createdResults_ = new();
+  private readonly ConcurrentBag<ICollection<TaskCreationRequest>>  createdTasks_   = new();
+  private readonly ILogger                                          logger_;
+  private readonly ConcurrentBag<ICollection<string>>               notifiedResults_ = new();
+  private readonly IObjectStorage                                   objectStorage_;
+  private readonly IPushQueueStorage                                pushQueueStorage_;
+  private readonly ConcurrentBag<(string id, ReadOnlyMemory<byte>)> resultsData_ = new();
```
lemaitre-aneo, line 57:
```suggestion
  private readonly ConcurrentBag<ICollection<(string id, ReadOnlyMemory<byte>)>> resultsData_ = new();
```

Common/src/Pollster/Agent.cs
```diff
@@ -243,90 +260,91 @@ public async Task<ICollection<TaskCreationRequest>> SubmitTasks(ICollection<Task
...
+  /// <inheritdoc />
+  public Task<ICollection<Result>> CreateResults(string                                                                  token,
+                                                 IEnumerable<(ResultCreationRequest request, ReadOnlyMemory<byte> data)> requests,
+                                                 CancellationToken                                                       cancellationToken)
+  {
+    ThrowIfInvalidToken(token);
 
-    return createdTasks;
+    var results = new List<Result>();
+    foreach (var rc in requests)
```
lemaitre-aneo, line 277: You could most likely go with a LINQ query:

```c#
var results = requests.Select(rc => (
  new Result(rc.request.SessionId,
             Guid.NewGuid()
               .ToString(),
             rc.request.Name,
             taskData_.TaskId,
             "",
             ResultStatus.Created,
             new List<string>(),
             DateTime.UtcNow,
             0,
             Array.Empty<byte>(),
  rc.data))
  .AsICollection();
createdResults_.Add(results.ViewSelect(rc => rc.Item1));
resultsData_.Add(results.ViewSelect(rc => (rc.Item1.ResultId, rc.Item2));
```

Common/src/Pollster/Agent.cs
```diff
@@ -243,90 +260,91 @@ public async Task<ICollection<TaskCreationRequest>> SubmitTasks(ICollection<Task
...
+                              taskData_.TaskId,
+                              "",
+                              ResultStatus.Created,
+                              new List<string>(),
+                              DateTime.UtcNow,
+                              0,
+                              Array.Empty<byte>());
+
+      results.Add(result);
     }
```
lemaitre-aneo, line 332: You can use a LINQ query to avoid appending element by element into a List.

lemaitre-aneo (approved):

PR #901, opened 2025-03-26
feat: manual deletion option for results

aneojgurhem:
# Motivation

Finalize import of existing data from the object storage by allowing users to create Results where in which the associated data will not be deleted by ArmoniK. Combined with import, the existing data within the object storage can still be managed externally to ArmoniK.

# Description

- Add a new field in the database called ManualDeletion.
- Check the value of this field and remove input payload only if set to false.
- During session purge data, only delete results where manual deletion is set to false.

# Testing

Enrich purge data unit test to validate that a Result with manual deletion set to `true` is not deleted.

# Impact

As no index was added on the ManualDeletion field in the database, purge data can become slower. It should not affect ArmoniK much as it not a critical query and is not called often compared to orchestration related queries.

# Checklist


dbrasseur-aneo (dismissed):

dbrasseur-aneo (dismissed):

Adaptors/MongoDB/src/Table/DataModel/ResultDataModelMapping.cs
```diff
@@ -55,6 +55,8 @@ public ResultDataModelMapping()
                                                 .SetDefaultValue(0);
                                               cm.MapProperty(nameof(Result.OpaqueId))
                                                 .SetIsRequired(true);
+                                              cm.MapProperty(nameof(Result.ManualDeletion))
+                                                .SetIsRequired(true);
```
lemaitre-aneo, line 59:
```suggestion
                                              cm.MapProperty(nameof(Result.ManualDeletion))
                                                .SetIgnoreIfDefault(true)
                                                .SetDefaultValue(0);
```

.docs/content/0.installation/5.version-migration.md
```diff
@@ -49,3 +49,20 @@ db.Result.updateMany({},
...
+Here is the MongoDB query to add this field:
+
+```js
+db.Result.updateMany({},
+    [{
+        $addFields: {
+            ManualDeletion: false
+        }
+    }])
+```
```
lemaitre-aneo, line 68: You can completely remove this if you set the field in database optional.

```suggestion
```

lemaitre-aneo (approved):

PR #939, opened 2025-04-16
refactor: postpone message when task is processing elsewhere

lemaitre-aneo:
# Motivation

During task acquisition, if the agent detects that the task is processing elsewhere, it acknowledges the message with the reasonning that the other is still healthy and processing it. However, there are cases that we failed to detect where the other responds, but might crash at some point afterwards, leading to message loss.

# Description

Instead of acknowledging the message, we postpone it (release to the queue), but with a delay. After the delay, the message is released to the queue and another pod will try to acquire the task, and check again the current pod.
As a consequence, we check the status of the task on the pod at a regular interval until it either completes, or crashes/times out.

# Impact

This should avoid all message loss related to message duplication, transient queue failure, and the race conditions that could follow.
Performance impact should be rather low thanks to the delay (by default 2 minutes).

If a pod crashes during its delay, the message will be available in the queue, thus avoiding message loss.

# Checklist

### DIFF
--- a/Common/src/Pollster/TaskHandler.cs
+++ b/Common/src/Pollster/TaskHandler.cs
@@ -57,6 +57,7 @@ public sealed class TaskHandler : IAsyncDisposable
   private readonly FunctionExecutionMetrics<TaskHandler> functionExecutionMetrics_;
   private readonly CancellationTokenSource               lateCts_;
   private readonly ILogger                               logger_;
+  private readonly TimeSpan                              messageDuplicationDelay_;
   private readonly IQueueMessageHandler                  messageHandler_;
   private readonly string                                ownerPodId_;
   private readonly string                                ownerPodName_;
@@ -68,6 +69,7 @@ public sealed class TaskHandler : IAsyncDisposable
   private readonly string                                token_;
   private readonly IWorkerStreamHandler                  workerStreamHandler_;
   private          IAgent?                               agent_;
+  private          TimeSpan?                             delayMessage_;
   private          DateTime?                             fetchedDate_;
   private          Action?                               onDispose_;
   private          Output?                               output_;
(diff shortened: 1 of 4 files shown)


aneojgurhem (approved):

Common/src/Pollster/TaskHandler.cs
```diff
@@ -68,6 +69,7 @@ public sealed class TaskHandler : IAsyncDisposable
   private readonly string                                token_;
   private readonly IWorkerStreamHandler                  workerStreamHandler_;
   private          IAgent?                               agent_;
+  private          TimeSpan?                             delayMessage_;
```
jfonseca-aneo, line 72: this naming makes me think that this is a message an not a `TimeSpan`, i'd inverse to `messageDelay_`

Compute/PollingAgent/src/Program.cs
```diff
@@ -199,22 +198,9 @@ public static async Task<int> Main(string[] args)
                           });
 
       app.MapGet("/taskprocessing",
-                 async () =>
-                 {
-                   var pollster  = app.Services.GetRequiredService<Common.Pollster.Pollster>();
-                   var exManager = app.Services.GetRequiredService<ExceptionManager>();
-
-                   var check = await pollster.Check(HealthCheckTag.Liveness)
-                                             .ConfigureAwait(false);
```
jfonseca-aneo, line 208: why is this healthcheck no longer needed ?

Common/src/Pollster/TaskHandler.cs
lemaitre-aneo: It is a verb. It is the action of delaying a message.

Compute/PollingAgent/src/Program.cs
lemaitre-aneo: Because we recheck later if the pod is still there and processing the task or not (when the message is pulled again on another pod).

jfonseca-aneo (approved):

PR #941, opened 2025-04-18
fix: cancel retried subtask

lemaitre-aneo:
# Motivation

When the agent crashes during post-processing, subtasks might end-up created, but not in the queue. This adds noise when debugging where there are tasks in creating/submitted that will never progress, while nothing depend on them.

# Description

If a task is acquired in the processing state, we check if there are any subtasks in the pending state. If so, it means the task has finished, completed its results and crashed while submitting subtasks. In that case, we can finish submitting the subtasks on behalf of the original task, and complete it. Otherwise, we cancel subtasks and retry the task.

# Testing

Artificially stop the post-processing at any given step, and check the tasks end-up in the right status (either completed or retried).

# Impact

All cases where it is safe to complete the task on behalf of the other pod are handled. All the other cases are safe to cancel the subtasks and retry crashed task.

# Checklist


Adaptors/MongoDB/tests/BsonSerializerTest.cs
aneojgurhem, line 1: a test should be added in this file

Common/src/Storage/TaskLifeCycleHelper.cs
```diff
@@ -609,4 +610,256 @@ await taskTable.UpdateManyTasks(data => (data.Status == TaskStatus.Submitted ||
...
+                                                         "One of the dependent tasks timed out.",
+                                                         CancellationToken.None)
+                                   .ConfigureAwait(false);
+        break;
+      default:
+        throw new ArgumentOutOfRangeException();
+    }
+  }
+
+  public static async Task<TaskStatus> HandleTaskCrashedWhileProcessing(ITaskTable        taskTable,
```
aneojgurhem, line 783: doc

Common/src/Storage/TaskLifeCycleHelper.cs
```diff
@@ -609,4 +610,256 @@ await taskTable.UpdateManyTasks(data => (data.Status == TaskStatus.Submitted ||
                    .ConfigureAwait(false);
     return sessionData;
   }
+
+  public static async Task CompleteTaskAsync(ITaskTable        taskTable,
```
aneojgurhem, line 614: doc

aneojgurhem (approved):

PR #943, opened 2025-04-21
fix: Properly handle cancellation in worker RPC

lemaitre-aneo:
# Motivation

When a cancellation is triggered during a RPC, the RPC throws a `RpcException` instead of an `OperationCanceledException` as expected. Consequently, tasks were retried when the agent was shutting down instead of just being requeued.

# Description

The check to know whether we are stopping the agent now also checks `RpcException` when the `InnerException` is an `OperationCanceledException`. It also uses the correct cancellationToken to verify.

# Testing

This has been tested locally, and a Unit test has been added to ensure both `OperationCanceledException` and `RpcException` are detected.

# Impact

During scale-down, tasks will now just be requeued without incrementing the retry counter. This will avoid task errors just for being retried too many times while being retried on scaling down pods.

# Additional information

If a task somehow makes the pod unhealthy without generating any direct error, this change might make it retry indefinitely.

# Checklist

### DIFF
--- a/Common/src/Pollster/TaskHandler.cs
+++ b/Common/src/Pollster/TaskHandler.cs
@@ -1015,7 +1015,7 @@ public sealed class TaskHandler : IAsyncDisposable
                   };
       await HandleErrorResubmitAsync(e,
                                      taskData_,
-                                     earlyCts_.Token)
+                                     lateCts_.Token)
         .ConfigureAwait(false);
     }
   }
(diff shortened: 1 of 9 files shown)


Common/src/Pollster/TaskHandler.cs
```diff
@@ -1015,7 +1015,7 @@ await agentHandler_.Stop(lateCts_.Token)
                   };
       await HandleErrorResubmitAsync(e,
                                      taskData_,
-                                     earlyCts_.Token)
+                                     lateCts_.Token)
```
jfonseca-aneo, line 1018: I fail to see how this change relates to the PR description

lemaitre-aneo: To check if the cancellation has been triggered, we need to check the proper cancellation token. During task execution, we use the late cancellation token, so we need to check the same.

PR #948, opened 2025-05-13
refactor:   move partitionId handling to pollster interface

ereali-aneo:
# Motivation

Originally, the `partitionId` was passed individually to each queue implementation. With the introduction of a new multi-queue system, this approach became awkward and redundant. Managing the same variable across all queues introduced unnecessary duplication and complexity.

This change aims to improve consistency by moving the `partitionId` handling to the pollster interface, which is more logically suited to manage partition context.

# Description

This PR includes:

* A change to the `IPollingQueueStorage` interface to include the `partitionId` parameter.
* Refactoring of all queue implementations to remove internal handling of `partitionId`.
* Update of the Terraform compute plane configuration to send the `partitionId` to the pollster instead of to each queue.
* Adjustments to existing tests to ensure they still build and run with the new interface.

Note: Tests were adjusted for compatibility but not rewritten or thoroughly adapted to validate the change end-to-end.

# Testing

* Unit and integration tests were executed via the CI pipeline to ensure that the system builds and behaves as expected.

# Impact

* **Breaking change**: Consumers of the polling queue interface must now pass the `partitionId` explicitly.
* **Improved modularity**: Queue implementations are now simpler and stateless with respect to partitions.
* 
# Checklist

* [x] My code adheres to the coding and style guidelines of the project.
* [x] I have performed a self-review of my code.
* [x] I have commented my code, particularly in hard-to-understand areas.
* [x] I have made corresponding changes to the documentation.
* [ ] I have thoroughly tested my modifications and added tests when necessary.
* [x] Tests pass locally and in the CI.
* [ ] I have assessed the performance impact of my modifications.


Adaptors/Amqp/src/PullQueueStorage.cs
```diff
@@ -69,46 +65,52 @@ public override async Task Init(CancellationToken cancellationToken)
...
+                                                        .Select(i => new AsyncLazy<ISenderLink>(async () => new SenderLink(new Session(await ConnectionAmqp
+                                                                                                                                             .GetConnectionAsync()
+                                                                                                                                             .ConfigureAwait(false)),
+                                                                                                                           $"{partitionId}###SenderLink{i}",
+                                                                                                                           $"{partitionId}###q{i}")))
+                                                        .ToArray());
+    var senders = await sendersArray.Select(lazy => lazy.Value)
+                                    .WhenAll();
+    var receivers = await receiversArray.Select(lazy => lazy.Value)
+                                        .WhenAll();
```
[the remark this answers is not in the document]
Nico-dl05: I think, it's ok but it might need an another validation

Nico-dl05 (approved):
Need to checkout precisely the lazy part. Whether it's useful or not calling lazy

Adaptors/Amqp/src/PullQueueStorage.cs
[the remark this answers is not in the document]
lemaitre-aneo: It is safe to eagerly await them all at this point, but I would advise against it, as it is not needed and just end up consuming more resources

Adaptors/Amqp/src/PullQueueStorage.cs
```diff
@@ -69,46 +65,52 @@ public override async Task Init(CancellationToken cancellationToken)
...
+                                                                        int                                        nbMessages,
                                                                         [EnumeratorCancellation] CancellationToken cancellationToken = default)
   {
     var nbPulledMessage = 0;
 
+    if (string.IsNullOrEmpty(partitionId))
+    {
+      throw new ArgumentException($"{nameof(partitionId)} is not defined.");
+    }
+
```
aneojgurhem, line 83: not needed here. It should be added in the Pollster so that it is not needed in every plugin.

Adaptors/RabbitMQ/src/PullQueueStorage.cs
```diff
@@ -59,6 +51,26 @@ public override async Task Init(CancellationToken cancellationToken)
...
+  }
+
+  public async IAsyncEnumerable<IQueueMessageHandler> PullMessagesAsync(string                                     partitionId,
+                                                                        int                                        nbMessages,
+                                                                        [EnumeratorCancellation] CancellationToken cancellationToken = default)
+  {
+    if (string.IsNullOrEmpty(partitionId))
+    {
+      throw new ArgumentException($"{nameof(partitionId)} is not defined.");
+    }
```
aneojgurhem, line 65: not needed here

Adaptors/SQS/src/PullQueueStorage.cs
```diff
@@ -57,14 +57,33 @@ public PullQueueStorage(AmazonSQSClient           client,
...
-  public async IAsyncEnumerable<IQueueMessageHandler> PullMessagesAsync(int                                        nbMessages,
+  public async IAsyncEnumerable<IQueueMessageHandler> PullMessagesAsync(string                                     partitionId,
+                                                                        int                                        nbMessages,
                                                                         [EnumeratorCancellation] CancellationToken cancellationToken)
   {
+    if (string.IsNullOrEmpty(partitionId))
+    {
+      throw new ArgumentException($"{nameof(partitionId)} is not defined.");
+    }
+
```
aneojgurhem, line 68: not needed here

Base/tests/Queue/QueueStorageTests.cs
aneojgurhem, line 1: I expect a new PR (or in this one depending on the timing) to add tests for SQS and PubSub here! I think there is a bit of logic to rework in the setup of these tests too.

aneojgurhem (changes requested):

Base/tests/Queue/QueueStorageTests.cs
ereali-aneo: I think this could be better to do a new one for readability and timing .

aneojgurhem (approved):

PR #1081, opened 2025-09-30
fix: retry message deadline extension in SQS adapter

aneojgurhem:
# Motivation

Improve the likelihood to extend on flight message deadline when there is an issue. This reduce the probability of a message representing a task in Processing being redistributed while still being processed elsewhere by an agent that was not able to extend the deadline for some reason.

# Description

Add a loop to retry message deadline extension.

# Testing

- Pipelines are passing.
- More extensive testing will be made later on where a few more fixes will be implemented.

# Impact

- Improve ArmoniK stability during connection issues with SQS. For instance, DNS issues that impede acces to AWS services.


aneojgurhem:
yes in another PR

PR #1193, opened 2026-04-24
refactor: make AMQP adaptor compatible with latest RabbitMQ

aneojgurhem:
# Motivation

RabbitMQ 4.x requires AMQP 1.0 queue addresses to be prefixed with `/queues/` and queues to be pre-declared before use. The previous code used a hardcoded `###` separator in queue and link names, which is not valid AMQP 1.0 addressing and broke with the latest RabbitMQ image.

# Description

**AMQP adapter (`Adaptors/Amqp`, `Adaptors/QueueCommon`)**
- Add a `Prefix` property to `Amqp` options (defaults to `""`): prepended to all queue addresses and link names.
- Add a `Separator` property to `Amqp` options (defaults to `"###"`): inserted between the partition ID and the queue/link index in AMQP resource names, replacing the previously hardcoded `###`.
- Use `prefix_` and `separator_` in `PullQueueStorage` and `PushQueueStorage` when building queue addresses and link names.

**Terraform — queue modules**
- Replace the single `partition` string field in `queue_envs` with a dedicated `queue_list` variable (`list(string)`) in every queue module (activemq, artemis, nats, pubsub, rabbitmq, sqs, none).
- Derive `queue_list` in the root locals from the partition configuration and pass it to every queue module call.

**Terraform — RabbitMQ module**
- Generate a `definitions.json` from `queue_list` at plan time: queues (`{partition}q{priority_band}`), a direct exchange, bindings, vhosts, permissions, and a properly SHA-256-hashed guest credential (salt generated once via `random_id`, hash computed via `data "external"` + Python 3).
- Upload `definitions.json` into the container and configure `rabbitmq.conf` to import it at startup (`definitions.import_backend = local_filesystem`).
- Add `Amqp__Prefix = /queues/` to the generated environment variables.
- Add `hashicorp/random` and `hashicorp/external` provider requirements.

# Testing

Validated by running the full terraform deployment (`up` / `destroy`) locally against RabbitMQ 4. The adaptor successfully pushes and pulls messages across all configured partitions.

# Impact

- **Breaking for existing state**: queue and link names change (the hardcoded `###` separator is now configurable and the prefix is added). A `terraform destroy` + re-apply is required for existing environments.
- **New providers**: `hashicorp/random` and `hashicorp/external` are required — run `terraform init -upgrade` after pulling this change.
- **Python 3 required** on the machine running terraform (used to compute the RabbitMQ password hash).
- No impact on non-RabbitMQ queue backends (activemq, artemis, nats, pubsub, sqs, none) beyond the `queue_list` input being added. The default `Separator = "###"` preserves the previous naming scheme for those backends.

# Additional Information

The number of physical RabbitMQ queues per partition scales automatically with `max_priority`: `nb_links = floor((max_priority - 1) / 10) + 1`. With the default `max_priority = 10`, each partition gets one queue (`q0`).

### DIFF
--- a/Adaptors/Amqp/src/PullQueueStorage.cs
+++ b/Adaptors/Amqp/src/PullQueueStorage.cs
@@ -44,6 +44,7 @@ public class PullQueueStorage : QueueStorage, IPullQueueStorage
 
   private readonly ConcurrentDictionary<string, AsyncLazy<IReceiverLink>[]> receivers_;
   private readonly ConcurrentDictionary<string, AsyncLazy<ISenderLink>[]>   senders_;
+  private readonly string                                                   separator_;
 
   public PullQueueStorage(QueueCommon.Amqp          options,
                           IConnectionAmqp           connectionAmqp,
@@ -55,6 +56,7 @@ public class PullQueueStorage : QueueStorage, IPullQueueStorage
     receivers_ = new ConcurrentDictionary<string, AsyncLazy<IReceiverLink>[]>();
     senders_   = new ConcurrentDictionary<string, AsyncLazy<ISenderLink>[]>();
     prefix_    = options.Prefix;
+    separator_ = options.Separator;
   }
 
   public override Task<HealthCheckResult> Check(HealthCheckTag tag)
@@ -96,8 +98,8 @@ public class PullQueueStorage : QueueStorage, IPullQueueStorage
                                                         .Select(i => new AsyncLazy<ISenderLink>(async () => new SenderLink(new Session(await ConnectionAmqp
                                                                                                                                              .GetConnectionAsync()
                                                                                                                                              .ConfigureAwait(false)),
-                                                                                                                           $"{prefix_}{partitionId}SenderLink{i}",
-                                                                                                                           $"{prefix_}{partitionId}q{i}")))
+                                                                                                                           $"{prefix_}{partitionId}{separator_}SenderLink{i}",
+                                                                                                                           $"{prefix_}{partitionId}{separator_}q{i}")))
                                                         .ToArray());
     while (nbPulledMessage < nbMessages)
     {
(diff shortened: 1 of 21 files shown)


Adaptors/Amqp/src/PullQueueStorage.cs
```diff
@@ -174,8 +176,8 @@ private AsyncLazy<IReceiverLink> CreateReceiver(string          partitionId,
            {
              var rl = new ReceiverLink(new Session(await connection.GetConnectionAsync()
                                                                    .ConfigureAwait(false)),
-                                       $"{partitionId}###ReceiverLink{link}",
-                                       $"{partitionId}###q{link}");
+                                       $"{prefix_}{partitionId}ReceiverLink{link}",
+                                       $"{prefix_}{partitionId}q{link}");
```
aneojgurhem: add separator variable

terraform/modules/storage/queue/rabbitmq/passwords.tf
```diff
@@ -0,0 +1,15 @@
+resource "random_id" "rabbit_password_salt" {
+  byte_length = 4
+}
+
+# RabbitMQ SHA256 hash: base64(salt + SHA256(salt + password))
+data "external" "rabbit_guest_password_hash" {
+  program = [
+    "python3", "-c",
+    "import hashlib,base64,json,sys;d=json.load(sys.stdin);s=bytes.fromhex(d['s']);p=d['p'].encode();print(json.dumps({'hash':base64.b64encode(s+hashlib.sha256(s+p).digest()).decode()}))"
```
AncientPatata, line 9: I guess [this isn't avoidable;](https://www.rabbitmq.com/docs/definitions#import-on-boot-nuances).. would be better to put it in its own file though I think.

AncientPatata (approved):

jfonseca-aneo (approved):

PR #1241, opened 2026-06-11
fix(storage): validate task dependencies before finalization

aneojgurhem:
# Motivation

When a task was submitted with dependencies referencing results that did not exist or that were in a terminal invalid state (`Aborted`, `DeletedData`), the system had no early detection mechanism. Tasks would end up permanently stuck in `Pending` state waiting for a dependency that can never become `Completed`.

# Description

Added two validation checks in `PrepareTaskDependencies` (called by `FinalizeTaskCreation`):

1. **Missing dependency detection** — after fetching all dependency results in a single query, the count is compared against the set of requested dependency IDs. If any are missing, the existing `ResultNotFoundException` is raised immediately with the list of unknown IDs.

2. **Invalid status detection** — a `switch` on each result's status routes `Completed` to the already-resolved set, `Created` to the pending-tracking set, and everything else (`Aborted`, `DeletedData`, `Unspecified`, or any unknown future status) to an invalid list. If the invalid list is non-empty after the loop, the new `ResultInvalidStatusException` is raised.

The two gRPC services that expose task submission (`GrpcSubmitterService`, `GrpcTasksService`) now catch both exception types and convert them to proper gRPC status codes:
- `ResultNotFoundException` → `StatusCode.NotFound`
- `ResultInvalidStatusException` → `StatusCode.FailedPrecondition`

As a side effect, the previous two-pass approach (one streaming query for completed-only, plus an implicit pass-through for everything else) is replaced by a single batched query whose in-memory result is reused for both validation and completed-dependency filtering.

### Files changed
- `Common/src/Exceptions/ResultInvalidStatusException.cs` *(new)* — thrown when a dependency is in a terminal non-completable state
- `Common/src/Storage/TaskLifeCycleHelper.cs` — validation logic using the new and existing exceptions
- `Common/src/gRPC/Services/GrpcSubmitterService.cs` — gRPC error mapping for both new exception paths
- `Common/src/gRPC/Services/GrpcTasksService.cs` — same
- `Common/tests/TaskLifeCycleHelperTest.cs` — test coverage

# Testing

Added a single parameterized NUnit test `TaskCreationWithInvalidDependencyShouldThrow` with three `[TestCase]` entries:

| `dependencyStatus` | Expected exception |
|---|---|
| `null` (result never inserted) | `ResultNotFoundException` |
| `ResultStatus.Aborted` | `ResultInvalidStatusException` |
| `ResultStatus.DeletedData` | `ResultInvalidStatusException` |

All three cases pass locally.

# Impact

- Task submission will now fail fast at `FinalizeTaskCreation` time rather than silently leaving tasks in `Pending` forever when dependencies are missing or in a terminal invalid state.
- Callers receive a typed gRPC error (`NOT_FOUND` or `FAILED_PRECONDITION`) instead of a silent hang.
- No performance regression: two async streaming queries replaced by one batched query with in-memory filtering.

# Additional Information

`ResultInvalidStatusException` also catches `Unspecified` and any unrecognised future status values via the `default` branch of the switch, making the validation forward-compatible.

### DIFF
--- a/Common/src/Storage/TaskLifeCycleHelper.cs
+++ b/Common/src/Storage/TaskLifeCycleHelper.cs
@@ -329,6 +329,7 @@ public static class TaskLifeCycleHelper
 
     var allDependencies       = new HashSet<string>();
     var completedDependencies = new List<string>();
+    var invalidDependencies   = new List<string>();
 
     // Get all the results that are a dependency of at least one task
     foreach (var request in taskRequests)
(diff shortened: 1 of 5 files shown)


Common/src/Storage/TaskLifeCycleHelper.cs
```diff
@@ -340,11 +340,43 @@ private static async Task<ICollection<MessageData>> PrepareTaskDependencies(ITas
...
+    {
+      var existingResultIds = results.Select(r => r.ResultId)
+                                     .ToHashSet();
+      var missingResultIds = allDependencies.Where(id => !existingResultIds.Contains(id));
+      logger.LogError("The following dependencies do not exist in the ResultTable: {@MissingResultIds}",
+                      missingResultIds);
+      throw new DependencyNotFoundException($"Some dependencies do not exist in the ResultTable: {string.Join(", ", missingResultIds)}");
+    }
+
+    // Aborted and DeletedData are terminal states — a result in these states can never become Completed,
```
lemaitre-aneo:
```suggestion
    // Aborted and DeletedData are terminal states: a result in these states can never become Completed,
```

Common/src/Storage/TaskLifeCycleHelper.cs
```diff
@@ -340,11 +340,43 @@ private static async Task<ICollection<MessageData>> PrepareTaskDependencies(ITas
...
+    // so any task depending on them would block forever
+    var invalidResults = results.Where(result => result.Status is ResultStatus.Aborted or ResultStatus.DeletedData)
+                                .AsICollection();
```
lemaitre-aneo: I think it would be better to iterate twice results in case of error, than materializing the filtered view always.
Actually, I think it would be better to merge the filter with the following loop, and have the filter within the loop itself.

lemaitre-aneo (dismissed):

Common/src/gRPC/Services/GrpcSubmitterService.cs
```diff
@@ -308,6 +308,20 @@ await resultTable_.DeleteResults(outputs,
...
+                         "Error while submitting tasks due to invalid dependency status");
+      throw new RpcException(new Status(StatusCode.FailedPrecondition,
+                                        "One or more dependencies have an invalid status"));
+    }
+    catch (ResultNotFoundException e)
+    {
+      logger_.LogWarning(e,
+                         "Error while submitting tasks due to dependency that does not exist");
+      throw new RpcException(new Status(StatusCode.NotFound,
+                                        "One or more dependencies does not exist"));
```
jfonseca-aneo:
```suggestion
                                        "One or more dependencies do not exist"));
```

Common/src/gRPC/Services/GrpcTasksService.cs
```diff
@@ -571,5 +571,19 @@ await TaskLifeCycleHelper.DeleteTasksAsync(taskTable_,
...
+                         "Error while submitting tasks due to invalid dependency status");
+      throw new RpcException(new Status(StatusCode.FailedPrecondition,
+                                        "One or more dependencies have an invalid status"));
+    }
+    catch (ResultNotFoundException e)
+    {
+      logger_.LogWarning(e,
+                         "Error while submitting tasks due to dependency that does not exist");
+      throw new RpcException(new Status(StatusCode.NotFound,
+                                        "One or more dependencies does not exist"));
```
jfonseca-aneo:
```suggestion
                                        "One or more dependencies do not exist"));
```

jfonseca-aneo:
Should't the introduced `ResultInvalidStatusException` also be used in the `CreateLargeTasks` path ? Or  the situation this exception aims to catch is not possible in this case ?

jfonseca-aneo (approved):

PR #1244, opened 2026-06-15
refactor(grpc): bind ExceptionInterceptor to ExceptionManager

lemaitre-aneo:
# Motivation

`ExceptionInterceptor` used to maintain its own error queue and `MaxErrorAllowed`
threshold, in parallel with the application-wide `ExceptionManager` that
already tracks errors and drives shutdown. Two independent error-management
systems made it harder to reason about overall service health and duplicated
logic.

# Description

Bind `ExceptionInterceptor` to `ExceptionManager` so there is a single source
of truth for error accounting.

- Constructor now takes an `ExceptionManager` instead of
  `Injection.Options.Submitter`. Errors that survive the client-error
  short-circuit are reported via `ExceptionManager.RecordError`, and
  successful calls are reported via `ExceptionManager.RecordSuccess`.
- The `IHealthCheckProvider.Check` implementation reports `Unhealthy` only
  when `tag == HealthCheckTag.Readiness` **and** the manager is in `Failed`
  state. All other combinations, notably `Liveness`, return `Healthy`,
  matching the readiness-vs-liveness convention used elsewhere.
- The local concurrent error queue and `maxAllowedErrors_` counter are
  removed.
- `ExceptionInterceptorTests` are updated to construct a real
  `ExceptionManager` (via a small `BuildExceptionManager(int maxError)` helper
  modeled on `ExceptionManagerTests`) and to query
  `HealthCheckTag.Readiness` instead of `Liveness`. The test fixture now
  manages an `ApplicationLifetime` in `SetUp`/`TearDown`.

The DI wiring in `Control/Submitter/src/Program.cs` already provides an
`ExceptionManager` (registered via `AddExceptionManager`), so the new
constructor is satisfied without further changes.

# Testing

- `dotnet build Common/tests/ArmoniK.Core.Common.Tests.csproj`: 0 errors.
- `dotnet test ... --filter FullyQualifiedName~ExceptionInterceptorTests`:
  3 passed, 0 failed (Unary, ClientStream, ServerStream variants).

The existing call sequences (4 successes, 4 client errors, then server errors
until the budget is exhausted) still drive the interceptor to Unhealthy at
exactly the same call number as before, because:

- Client errors (RpcException with `NotFound` / `InvalidArgument` /
  `Cancelled` / …) still short-circuit `HandleException` without recording.
- Server errors call `RecordError`; the (n+1)-th flips
  `ExceptionManager.Failed` to true, matching the old
  `errors_.Count > maxAllowedErrors_` trigger point.
- `RecordSuccess` is a no-op when `nbError == 0` or when `Failed` is already
  true, so success-after-failure does not un-fail the manager (matching the
  old non-decrementing queue).

# Impact

- Behaviour change: `ExceptionInterceptor` only reports `Unhealthy` on the
  Readiness health-check tag now (previously, on every tag). Any consumer
  polling Liveness for "too many errors" will no longer see Unhealthy, but
  that signal was conceptually a readiness signal anyway, and Liveness should
  remain Healthy while the process is still running.
- Configuration: the interceptor no longer reads
  `Submitter.MaxErrorAllowed`; the equivalent limit is now configured on the
  shared `ExceptionManager.Options.MaxError` value used by the rest of the
  service.
- No new dependencies. No public-API change beyond the constructor of
  `ExceptionInterceptor`, which is resolved via DI.

# Additional Information

The interceptor's `[Obsolete]` test methods stay obsolete, they still
exercise the obsolete `ISubmitter` surface, unrelated to this change.

# Checklist

### DIFF
--- a/Common/src/gRPC/ExceptionInterceptor.cs
+++ b/Common/src/gRPC/ExceptionInterceptor.cs
@@ -1,5 +1,5 @@
 // This file is part of the ArmoniK project
-// 
+//
 // Copyright (C) ANEO, 2021-2026. All rights reserved.
 // 
 // This program is free software: you can redistribute it and/or modify
@@ -16,12 +16,11 @@
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 using System;
-using System.Collections.Concurrent;
 using System.Threading.Tasks;
 
 using ArmoniK.Core.Base;
 using ArmoniK.Core.Base.DataStructures;
-using ArmoniK.Core.Common.Injection.Options;
+using ArmoniK.Core.Common.Utils;
 
 using Grpc.Core;
 using Grpc.Core.Interceptors;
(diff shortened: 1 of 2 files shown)


Common/src/gRPC/ExceptionInterceptor.cs
```diff
@@ -37,48 +36,40 @@ namespace ArmoniK.Core.Common.gRPC;
...
-                     errors_.Count,
-                     maxAllowedErrors_);
-    return Task.FromResult(maxAllowedErrors_ < 0 || errors_.Count <= maxAllowedErrors_
-                             ? HealthCheckResult.Healthy()
-                             : HealthCheckResult.Unhealthy("Too many errors recorded",
-                                                           new AggregateException(errors_)));
-  }
+    => Task.FromResult((tag, exceptionManager_.Failed) switch
+                       {
+                         (HealthCheckTag.Readiness, true) => HealthCheckResult.Unhealthy("Too many error recorded, application is shutting down"),
```
aneojgurhem: I think a small comment about why it is readiness would be welcomed

aneojgurhem (changes requested):

Common/src/gRPC/ExceptionInterceptor.cs
```diff
@@ -37,48 +36,42 @@ namespace ArmoniK.Core.Common.gRPC;
...
-    return Task.FromResult(maxAllowedErrors_ < 0 || errors_.Count <= maxAllowedErrors_
-                             ? HealthCheckResult.Healthy()
-                             : HealthCheckResult.Unhealthy("Too many errors recorded",
-                                                           new AggregateException(errors_)));
-  }
+    => Task.FromResult((tag, exceptionManager_.Failed) switch
+                       {
+                         // If there is too many errors, the pod is marked as not ready to avoid Kubernetes sending new requests to the controller.
+                         // Liveness Unhealthy is not needed as the pod is killing itself, without the intervention of Kubernetes.
+                         (HealthCheckTag.Readiness, true) => HealthCheckResult.Unhealthy("Too many error recorded, application is shutting down"),
```
jfonseca-aneo:
```suggestion
                         (HealthCheckTag.Readiness, true) => HealthCheckResult.Unhealthy("Too many errors recorded, application is shutting down"),
```

jfonseca-aneo:
I think the change on how the errors are recorded is worth to mention in the PR description

lemaitre-aneo:
> I think the change on how the errors are recorded is worth to mention in the PR description

I don't understand your point: this is already documented in the PR description in the description and the impact sections.

aneojgurhem (approved):

jfonseca-aneo:
> > I think the change on how the errors are recorded is worth to mention in the PR description
> 
> I don't understand your point: this is already documented in the PR description in the description and the impact sections.

I mean that  the PR description mentions the post-failure case but not the pre-failure sliding-window

jfonseca-aneo (approved):

PR #1253, opened 2026-06-19
feat(submitter): add configurable grace delay

aneojgurhem:
# Motivation

The Submitter service was passing `TimeSpan.Zero` as the grace delay to `ExceptionManager`, meaning there was no delay between the early and late cancellation tokens on shutdown. Under load, this can cause in-flight work to be aborted before it completes. Making the delay configurable allows operators to tune shutdown behaviour to their workload.

# Description

- Added a new `GraceDelay` property (default: 30 seconds) to `Common/src/Injection/Options/Submitter.cs`.
- Wired it into `ExceptionManager.Options` in `Control/Submitter/src/Program.cs`, replacing the hardcoded `TimeSpan.Zero`.
- The option is read from the existing `Submitter` configuration section, so no new configuration section is required.

# Testing

No new automated tests added — the change is a straightforward options binding. The default value (30 s) provides a reasonable grace window and the path can be exercised manually by configuring a short delay and observing graceful drain behaviour.

# Impact

- **Configuration**: operators can now set `Submitter:GraceDelay` (e.g. `"00:01:00"`) to tune the window between early and late cancellation on shutdown.
- **Behaviour**: the grace delay changes from the previous hardcoded `TimeSpan.Zero` (no delay) to 30 seconds by default, giving in-flight requests time to complete before the late cancellation fires.
- **Performance**: no impact on steady-state throughput.

# Additional Information

None.

# Checklist

### DIFF
--- a/Control/Submitter/src/Program.cs
+++ b/Control/Submitter/src/Program.cs
@@ -86,6 +86,9 @@ public static class Program
     {
       AppDomain.CurrentDomain.AssemblyResolve += new CollocatedAssemblyResolver(logger.GetLogger()).AssemblyResolve;
 
+      var submitterOptions = builder.Configuration.GetSection(Common.Injection.Options.Submitter.SettingSection)
+                                    .Get<Common.Injection.Options.Submitter>() ?? new Common.Injection.Options.Submitter();
+
       builder.Host.UseSerilog(logger.GetSerilogConf());
 
       builder.Services.AddLogging(logger.Configure)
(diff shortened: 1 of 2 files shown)


Control/Submitter/src/Program.cs
```diff
@@ -105,7 +108,7 @@ public static async Task<int> Main(string[] args)
              .AddInitializedOption<InitServices>(builder.Configuration,
                                                  InitServices.SettingSection)
              .AddSingleton<InitDatabase>()
-             .AddExceptionManager(sp => new ExceptionManager.Options(TimeSpan.Zero,
+             .AddExceptionManager(sp => new ExceptionManager.Options(submitterOptions.GraceDelay,
```
lemaitre-aneo, line 111: You should fetch the submitter options from sp instead of loading it twice

lemaitre-aneo (commented):
In retrospective, I think there is the "early exit during grace delay" missing: If the grace delay fires, it should wait only for the current requests to be processed, and not wait the full grace delay.
I think adding a counter inside the interceptor would be a good way to integrate this mechanism.

PR #1257, opened 2026-06-23
feat(postgres): add PostgreSQL adapter

aneojgurhem:
# Motivation

ArmoniK.Core currently supports MongoDB as its primary database backend. This PR adds PostgreSQL as a second, fully-supported backend, giving operators a widely-available, open-source relational option with strong ACID guarantees and native support for change notification via logical replication (used for the task watcher).

# Description

Introduces a new adapter under `Adaptors/PostgresSQL/` implementing all core storage interfaces (`ITaskTable`, `IResultTable`, `ISessionTable`, `IPartitionTable`, `ITaskWatcher`) against PostgreSQL using raw ADO.NET + Npgsql, plus Terraform and CI support to deploy and test it alongside MongoDB.

**Core implementation**
- Schema and DI: tables created on startup via `DatabaseInitializer`; services registered through the standard `IDependencyInjectionBuildable` pattern.
- Bulk inserts via `NpgsqlBatch` for `CreateTasks`, `CreateResults`, etc.
- Task watcher built on PostgreSQL logical replication (WAL) via `NpgsqlReplicationConnection`, with a single WAL replication connection shared across concurrent event subscribers.
- Binary WAL mode with single-pass typed tuple reading — watcher data is read directly from WAL tuples, no secondary query needed. `REPLICA IDENTITY FULL` was dropped in favour of tolerating spurious watcher events.
- `ExpressionToSql` translates `Expression<Func<T, bool>>` trees to parameterized SQL; all values go through `AddWithValue`, never string interpolation.
- Justfile gained a `database` variable (`mongo` | `postgresql`, default `mongo`) and Terraform provisions the corresponding storage backend.
- CI: `Adaptors/PostgresSQL/tests` added to the standard unit-test matrices, and the Stream/HtcMock integration pipelines now run against both `mongo` and `postgresql`.

**Correctness and concurrency fixes made while hardening the adapter**
- SQL injection prevention: parameterized sort-key handling (escaping instead of an allowlist) and LIKE wildcard escaping (`%`, `_`, `\`).
- Race conditions: `FOR UPDATE` locking on session pre-read and on concurrent dependency removal (serializes two transactions that each clear one of N leaf dependencies on the same aggregation task, preventing it from being stranded in Pending); a TOCTOU fix in `UpdateOneResult`.
- WAL broadcaster: three concurrency bugs fixed, WAL acknowledged after write (not before), subscription to WAL happens before the DB snapshot to eliminate an event-loss window, and the `slotReady_` completion source is resolved correctly if the WAL loop exits before slot creation; a lock-release-before-await fix in `EnsureStartedAndRegister`.
- Query correctness: empty `ORDER BY` handled in `ListApplicationsAsync`, `Equals(null)` translated to `IS NULL`, null dictionary-key guard in JSONB deserialization, `string.IsNullOrEmpty` support in `ExpressionToSql`, `RemainingDataDependencies` populated correctly in list/find.
- Connection handling: `MaxPoolSize` applied when `ConnectionString` is set, `max_connections` raised to 1000, `AppContext.SetSwitch` promoted to a static initializer (was previously called per-connection), locale fixed to `C` for the embedded test server, unused Dapper dependency removed.
- The same empty-sort-fields bug found in `ListApplicationsAsync` also existed in the MongoDB and in-memory adapters; both got the equivalent one-line fix for parity.

# Testing

- New `Adaptors/PostgresSQL/tests/` project with `TestBase` coverage shared with the other adapters, plus adapter-specific coverage for LIKE escaping, `FindOneAndUpdate`-style `before=true` snapshot semantics, `ORDER BY` column quoting, and regression tests for each fix above.
- Embedded PostgreSQL test setup was sped up and gained support for pointing at an external instance for CI.
- CI now runs the PostgreSQL adapter's unit tests and the Stream/HtcMock integration suites against a real PostgreSQL deployment, in addition to MongoDB.
- Existing MongoDB and Common test suites are unaffected by this change.

# Impact

- New dependency: `Npgsql` (Dapper, initially added, was unused and has been removed).
- No breaking changes to existing adapters or public interfaces.
- MongoDB remains the default backend; PostgreSQL is opt-in via `database=postgresql` at deploy time.
- Binary WAL mode, shared WAL connections, and `NpgsqlBatch` bulk inserts keep round-trips low for high-throughput workloads.

# Additional Information

- The adapter targets `net10.0`, consistent with the rest of the project.
- `REPLICA IDENTITY FULL` was dropped because it imposes a write-amplification cost on every UPDATE; the watcher tolerates spurious events by re-reading task state when a WAL event fires.

# Checklist


Control/Submitter/src/Program.cs
```diff
@@ -89,12 +92,16 @@ public static async Task<int> Main(string[] args)
...
-                                 logger.GetLogger())
-             .AddAdapter(builder.Configuration,
-                         nameof(Components.QueueAdaptorSettings),
-                         logger.GetLogger())
+             .AddHttpClient();
+
+      builder.Services.AddPostgresComponents(builder.Configuration,
+                                             logger.GetLogger());
+      builder.Services.AddMongoComponents(builder.Configuration,
+                                          logger.GetLogger());
```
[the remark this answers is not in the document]
jfonseca-aneo: fixed by https://github.com/aneoconsulting/ArmoniK.Core/pull/1257/commits/dedf6fa5f2875b25142f810c3c0fa2ecec3eaaee

Control/Metrics/src/Program.cs
```diff
@@ -62,10 +63,14 @@ public static async Task<int> Main(string[] args)
...
-      builder.Services.AddLogging(logger.Configure)
-             .AddMongoComponents(builder.Configuration,
-                                 logger.GetLogger())
-             .AddSingleton(builder.Configuration.GetInitializedValue<MetricsExporter>(MetricsExporter.SettingSection))
+      builder.Services.AddLogging(logger.Configure);
+
+      builder.Services.AddPostgresComponents(builder.Configuration,
+                                             logger.GetLogger());
+      builder.Services.AddMongoComponents(builder.Configuration,
+                                          logger.GetLogger());
```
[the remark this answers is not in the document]
jfonseca-aneo: fixed by https://github.com/aneoconsulting/ArmoniK.Core/commit/dedf6fa5f2875b25142f810c3c0fa2ecec3eaaee

Compute/PollingAgent/src/Program.cs
```diff
@@ -93,12 +94,16 @@ public static async Task<int> Main(string[] args)
...
-                                 logger.GetLogger())
-             .AddAdapter(builder.Configuration,
-                         nameof(Components.QueueAdaptorSettings),
-                         logger.GetLogger())
+             .AddArmoniKWorkerConnection(builder.Configuration);
+
+      builder.Services.AddPostgresComponents(builder.Configuration,
+                                             logger.GetLogger());
+      builder.Services.AddMongoComponents(builder.Configuration,
+                                          logger.GetLogger());
```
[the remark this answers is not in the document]
jfonseca-aneo: fixed by https://github.com/aneoconsulting/ArmoniK.Core/commit/dedf6fa5f2875b25142f810c3c0fa2ecec3eaaee

terraform/modules/storage/database/postgresql/main.tf
```diff
@@ -0,0 +1,66 @@
...
+}
+
+resource "docker_container" "database" {
+  name  = var.postgresql_params.database_name
+  image = docker_image.database.image_id
+
+  entrypoint = var.postgresql_params.ssl ? [
+    "bash", "-c",
+    "chmod 600 /postgresql-certificate/server.key && exec /usr/local/bin/docker-entrypoint.sh postgres -c ssl=on -c ssl_cert_file=/postgresql-certificate/server.crt -c ssl_key_file=/postgresql-certificate/server.key -c ssl_ca_file=/postgresql-certificate/ca.pem -c wal_level=logical -c max_connections=${var.postgresql_params.max_connections}"
+  ] : ["docker-entrypoint.sh", "postgres", "-c", "wal_level=logical", "-c", "max_connections=${var.postgresql_params.max_connections}"]
```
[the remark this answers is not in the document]
jfonseca-aneo: fixed in https://github.com/aneoconsulting/ArmoniK.Core/pull/1257/commits/57658b07624d0cec2aa05d92aa9f0d10fa282b95

Adaptors/PostgresSQL/src/Common/Migrations/Scripts/0001_InitialSchema.sql
```diff
@@ -0,0 +1,153 @@
...
+-- Results table
+CREATE TABLE IF NOT EXISTS results (
+    session_id      TEXT NOT NULL,
+    result_id       TEXT PRIMARY KEY,
+    name            TEXT NOT NULL DEFAULT '',
+    created_by      TEXT NOT NULL DEFAULT '',
+    completed_by    TEXT NOT NULL DEFAULT '',
+    owner_task_id   TEXT NOT NULL DEFAULT '',
+    status          INTEGER NOT NULL,
+    dependent_tasks TEXT[] NOT NULL DEFAULT '{}',
```
lemaitre-aneo: Is not needed thanks to the join table.

Adaptors/PostgresSQL/src/Common/NpgsqlConnectionProvider.cs
```diff
@@ -0,0 +1,317 @@
...
+  /// <summary>
+  ///   The Npgsql data source (connection pool)
+  /// </summary>
+  public NpgsqlDataSource DataSource { get; }
+
+  /// <inheritdoc />
+  public void Dispose()
+  {
+    DataSource.Dispose();
+    GC.SuppressFinalize(this);
```
lemaitre-aneo: Class does not have a finalizer.

Adaptors/PostgresSQL/src/Common/NpgsqlConnectionProvider.cs
```diff
@@ -0,0 +1,317 @@
...
+using Npgsql.Replication;
+
+using NpgsqlTypes;
+
+namespace ArmoniK.Core.Adapters.PostgresSQL.Common;
+
+/// <summary>
+///   Provides NpgsqlDataSource management, schema initialization, and health check
+/// </summary>
+public class NpgsqlConnectionProvider : IInitializable, IDisposable
```
lemaitre-aneo: DataSource is AsyncDisposble, so I would say this class should also be AsyncDisposable

Adaptors/PostgresSQL/src/Common/NpgsqlConnectionProvider.cs
```diff
@@ -0,0 +1,317 @@
...
+    // DbUp opens its own connections against the same database, so the advisory lock
+    // is held on a separate connection/transaction for the duration of the upgrade to
+    // serialise concurrent instances migrating at the same time.
+    await using var lockTransaction = await connection.BeginTransactionAsync(cancellationToken)
+                                                      .ConfigureAwait(false);
+
+    await using (var lockCmd = connection.CreateCommand())
+    {
+      lockCmd.Transaction = lockTransaction;
+      lockCmd.CommandText = "SELECT pg_advisory_xact_lock(7243658712345678)";
```
lemaitre-aneo, line 118: This command requires some explanation...

Adaptors/PostgresSQL/src/Options/TableStorage.cs
lemaitre-aneo, line 1: I think it is unnecessary for this design.

Adaptors/PostgresSQL/src/Common/NpgsqlConnectionProvider.cs
```diff
@@ -0,0 +1,317 @@
...
+  ///   Get a new connection from the pool
+  /// </summary>
+  /// <param name="cancellationToken">Cancellation token</param>
+  /// <returns>An open NpgsqlConnection</returns>
+  public async Task<NpgsqlConnection> GetConnectionAsync(CancellationToken cancellationToken = default)
+  {
+    var connection = await DataSource.OpenConnectionAsync(cancellationToken)
+                                     .ConfigureAwait(false);
+    return connection;
+  }
```
lemaitre-aneo:
```suggestion
  public ValueTask<NpgsqlConnection> GetConnectionAsync(CancellationToken cancellationToken = default)
  => DataSource.OpenConnectionAsync(cancellationToken));
```

Adaptors/PostgresSQL/src/PartitionTable.cs
```diff
@@ -0,0 +1,293 @@
...
+    => connectionProvider_ = connectionProvider;
+
+  /// <inheritdoc />
+  public async Task CreatePartitionsAsync(IEnumerable<PartitionData> partitions,
+                                          CancellationToken          cancellationToken = default)
+  {
+    await using var connection = await connectionProvider_.GetConnectionAsync(cancellationToken)
+                                                          .ConfigureAwait(false);
+    await using var transaction = await connection.BeginTransactionAsync(cancellationToken)
+                                                  .ConfigureAwait(false);
```
lemaitre-aneo: Do we really need a transaction for creating partitions?

Adaptors/PostgresSQL/src/ResultTable.cs
```diff
@@ -0,0 +1,541 @@
...
+  session_id, result_id, name, created_by, completed_by, owner_task_id,
+  status, dependent_tasks, creation_date, completion_date, size, opaque_id, manual_deletion
+) VALUES (
+  @session_id, @result_id, @name, @created_by, @completed_by, @owner_task_id,
+  @status, @dependent_tasks, @creation_date, @completion_date, @size, @opaque_id, @manual_deletion
+)");
+        AddResultInsertParameters(cmd.Parameters,
+                                  result);
+        batch.BatchCommands.Add(cmd);
+      }
```
lemaitre-aneo, line 91: I think it would be better to avoid batching commands, and instead insert many values in a single command.

Adaptors/PostgresSQL/src/ResultTable.cs
```diff
@@ -0,0 +1,541 @@
...
+    await transaction.CommitAsync(cancellationToken)
+                     .ConfigureAwait(false);
+
+    Logger.LogDebug("Created {count} results",
+                    results.Count);
+  }
+
+  /// <inheritdoc />
+  public async Task AddTaskDependencies(IDictionary<string, ICollection<string>> dependencies,
+                                        CancellationToken                        cancellationToken = default)
```
lemaitre-aneo, line 137: I am not sure why this needs to be implemented this way as there is the join table.

Adaptors/PostgresSQL/src/ResultTable.cs
```diff
@@ -0,0 +1,541 @@
...
+    await using var transaction = await connection.BeginTransactionAsync(cancellationToken)
+                                                  .ConfigureAwait(false);
+
+    // Read before update; FOR UPDATE acquires a row-level lock so concurrent
+    // UpdateOneResult calls on the same result are serialized within the transaction.
+    await using var readCmd = connection.CreateCommand();
+    readCmd.Transaction = transaction;
+    readCmd.CommandText = "SELECT * FROM results WHERE result_id = @result_id FOR UPDATE";
+    readCmd.Parameters.AddWithValue("result_id",
+                                    resultId);
```
lemaitre-aneo: I am pretty sure update returning is a better pattern.

Adaptors/PostgresSQL/src/SessionTable.cs
```diff
@@ -0,0 +1,349 @@
...
+
+    await using var cmd = connection.CreateCommand();
+    cmd.CommandText = $"SELECT * FROM sessions WHERE {whereSql}";
+    SqlHelper.AddExpressionParameters(cmd,
+                                      whereParams);
+
+    await using var reader = await cmd.ExecuteReaderAsync(cancellationToken)
+                                      .ConfigureAwait(false);
+
+    var compiled = selector.Compile();
```
lemaitre-aneo: I think it should be better to compile before starting the request, to avoid performing the request in case of an error.

Adaptors/PostgresSQL/src/SessionTable.cs
```diff
@@ -0,0 +1,349 @@
...
+
+    return (results, totalCount);
+  }
+
+  /// <inheritdoc />
+  public async Task<SessionData?> UpdateOneSessionAsync(string                               sessionId,
+                                                        Expression<Func<SessionData, bool>>? filter,
+                                                        UpdateDefinition<SessionData>        updates,
+                                                        bool                                 before,
+                                                        CancellationToken                    cancellationToken = default)
```
lemaitre-aneo, line 237: Could be implemented with an update returing

Adaptors/PostgresSQL/src/TaskTable.cs
```diff
@@ -0,0 +1,839 @@
...
+                                                  .ConfigureAwait(false);
+
+    try
+    {
+      await using var batch = new NpgsqlBatch(connection,
+                                              transaction);
+
+      foreach (var task in tasks)
+      {
+        var insertCmd = new NpgsqlBatchCommand(@"
```
lemaitre-aneo, line 82: I think it is better to insert many values in one command instead of batching

Adaptors/PostgresSQL/src/TaskTable.cs
```diff
@@ -0,0 +1,839 @@
...
+                                   });
+    }
+  }
+
+  /// <inheritdoc />
+  public async Task<TaskData?> UpdateOneTask(string                            taskId,
+                                             Expression<Func<TaskData, bool>>? filter,
+                                             UpdateDefinition<TaskData>        updates,
+                                             bool                              before,
+                                             CancellationToken                 cancellationToken = default)
```
lemaitre-aneo, line 411: I think it could be implemented using update returning

Adaptors/PostgresSQL/src/ResultTable.cs
aneojgurhem: We did consider a single multi-row `INSERT ... VALUES (...), (...), (...)` before landing on `NpgsqlBatch`. We went through two earlier iterations:

1. Per-row named placeholders (`VALUES (@p0_0, @p0_1, ...), (@p1_0, ...)`) — dropped because the generated SQL string grows with `rows × columns`, which gets unwieldy for large batches.
2. An `unnest`-based insert (one typed array per column, fixed-size SQL regardless of row count) — dropped in favor of `NpgsqlBatch` to avoid the added complexity of array-parameter binding.

There's also a practical performance angle: in production, a single `Create` call inserts at most a few hundred rows (task submissions are typically batched in the low hundreds). At that scale, the per-command planning overhead of N `NpgsqlBatchCommand`s vs. one query plan for a true multi-row insert is negligible — the round-trip is the actual bottleneck, and `NpgsqlBatch` already collapses that to one. A single multi-row/`unnest` insert would only start paying off at row counts well beyond what we see in practice.

aneojgurhem: ## Bulk insert / upsert into Postgres via Npgsql: design notes

Context: we need two distinct write paths against Postgres from C# using Npgsql
directly (no EF Core). This note summarizes the recommended approach for each.

### Path 1: inserts, batches of ~100 rows per request

`COPY` (via Npgsql's `NpgsqlBinaryImporter`, i.e. `BeginBinaryImport`) is the
fastest bulk method overall, but its setup cost (opening the COPY stream,
protocol negotiation) is only amortized on large batches (thousands to millions
of rows). At ~100 rows per request it brings no measurable gain and adds
complexity, so we should not use it here.

Recommended: a single round-trip per batch, either

- **`NpgsqlBatch`** (Npgsql 6+): groups several parameterized commands into one
  network round-trip, each command still preparable. Preferred for readability.
- or a **multi-row parameterized `INSERT`** (`VALUES (@id0, ...), (@id1, ...)`)
  if we stay on an older Npgsql version.

Caveats:
- Postgres caps parameters at **65535 per statement**. At 100 rows this leaves
  a wide margin, but any variable-size batching must chunk to stay under it.
- Never interpolate values into the SQL string (injection + plan cache misses).
  Only the `(@idN, ...)` structure is concatenated, never the data.
- Reuse connections through the Npgsql pool; do not open one connection per
  request manually.

### Path 2: upserts, small volumes

Use `INSERT ... ON CONFLICT DO UPDATE` directly. No temp table needed at this
scale (the "COPY into temp table + merge" pattern only pays off for large
upsert batches).

```sql
INSERT INTO <table> (...)
VALUES (...)
ON CONFLICT (<key>) DO UPDATE SET ...
```

### Shared pitfall

`timestamp` vs `timestamptz`: Npgsql maps `DateTime` according to its `Kind`.
Mismatches here are a common source of silent bugs. Set `NpgsqlDbType`
explicitly on every parameter; Npgsql does no implicit numeric/temporal
conversion (unlike SqlBulkCopy).

Adaptors/PostgresSQL/src/Common/NpgsqlConnectionProvider.cs
jfonseca-aneo: done in https://github.com/aneoconsulting/ArmoniK.Core/pull/1257/commits/c168bd2387b868a52d2bcba0818c9046b8ca06a4

Adaptors/PostgresSQL/src/Common/NpgsqlConnectionProvider.cs
```diff
@@ -0,0 +1,360 @@
...
+  {
+    if (disposed_)
+    {
+      return;
+    }
+
+    disposed_ = true;
+    DataSource.Dispose();
+    GC.SuppressFinalize(this);
+  }
```
lemaitre-aneo:
```suggestion
  /// <inheritdoc />
  public ValueTask DisposeAsync()
  => DataSource.DisposeAsync();

  /// <inheritdoc />
  public void Dispose()
  => DataSource.Dispose();
```

Adaptors/PostgresSQL/src/Common/NpgsqlConnectionProvider.cs
```diff
@@ -0,0 +1,360 @@
...
+  ~NpgsqlConnectionProvider()
+  {
+    if (disposed_)
+    {
+      return;
+    }
+
+    disposed_ = true;
+    DataSource.Dispose();
+  }
```
lemaitre-aneo: The class has no unmanaged resources, so a finalizer is not needed. It is the implementation of DataSource that should have a finalizer to handle the case it has not been disposed.

Adaptors/PostgresSQL/src/ResultTable.cs
```diff
@@ -0,0 +1,672 @@
...
+      results.Add(RowMapper.MapToResult(reader));
+    }
+
+    await reader.CloseAsync()
+                .ConfigureAwait(false);
+
+    var dependentTasksByResultId = await LoadDependentTasksBatch(connection,
+                                                                 results,
+                                                                 cancellationToken)
+                                     .ConfigureAwait(false);
```
lemaitre-aneo: This should not be useful.

Adaptors/PostgresSQL/src/ResultTable.cs
```diff
@@ -0,0 +1,672 @@
...
+    // CTE selects it with FOR UPDATE (locking the row so concurrent UpdateOneResult
+    // calls on the same result are serialized), then the UPDATE runs against it.
+    await using var updateCmd = connection.CreateCommand();
+    updateCmd.Transaction = transaction;
+    var setClauses = SqlHelper.BuildSetClauses(updates,
+                                               updateCmd);
+    updateCmd.CommandText =
+      "WITH old AS (SELECT * FROM results WHERE result_id = @result_id FOR UPDATE) " +
+      $"UPDATE results SET {setClauses} FROM old WHERE results.result_id = old.result_id " +
+      "RETURNING old.*";
```
lemaitre-aneo:
```suggestion
      $"UPDATE results SET {setClauses} FROM old WHERE results.result_id = @result_id " +
      "RETURNING OLD.*";
```

We could easily target postgre 18: https://www.postgresql.org/docs/current/sql-update.html

Adaptors/PostgresSQL/src/ResultTable.cs
aneojgurhem: mettre une liste vide dans mongo et un test

Adaptors/PostgresSQL/src/ResultTable.cs
```diff
@@ -0,0 +1,672 @@
...
+      await using var cmd = connection.CreateCommand();
+      cmd.Transaction = transaction;
+      var setClauses = SqlHelper.BuildSetClauses(updates,
+                                                 cmd);
+      cmd.CommandText = $"UPDATE results SET {setClauses} WHERE {whereSql}";
+      SqlHelper.AddExpressionParameters(cmd,
+                                        whereParams);
+
+      totalMatched += await cmd.ExecuteNonQueryAsync(cancellationToken)
+                               .ConfigureAwait(false);
```
lemaitre-aneo: Can't you batch here?

Adaptors/PostgresSQL/tests/TaskTableTests.cs
```diff
@@ -0,0 +1,89 @@
...
+                                                       CancellationToken.None)
+                                       .ConfigureAwait(false);
+
+      Assert.That(tasks,
+                  Is.Empty);
+    }
+  }
+
+  [Test]
+  public async Task ListApplicationsAsyncShouldReturnEmptyForNegativePageSize()
```
lemaitre-aneo: Why is this test only for Postgres?

Adaptors/PostgresSQL/tests/TaskTableTests.cs
```diff
@@ -0,0 +1,89 @@
...
+  {
+    tableProvider_ = new PostgresDatabaseProvider();
+    var provider = tableProvider_.GetServiceProvider();
+
+    TaskTable = provider.GetRequiredService<ITaskTable>();
+    RunTests  = true;
+  }
+
+  [Test]
+  public async Task ListTasksAsyncShouldNotOverflowOffsetWithLargePaging()
```
lemaitre-aneo: I am not sure what's the purpose of this test. We will never have million-items pages. If it needs to be supported, it should go into the generic tests (and also test MongoDB), and if it is not needed, it should be removed entirely.

Adaptors/PostgresSQL/tests/ResultTableTests.cs
```diff
@@ -0,0 +1,117 @@
...
+  {
+    tableProvider_ = new PostgresDatabaseProvider();
+    var provider = tableProvider_.GetServiceProvider();
+
+    ResultTable = provider.GetRequiredService<IResultTable>();
+    RunTests    = true;
+  }
+
+  [Test]
+  public async Task AddTaskDependenciesTwiceShouldNotDuplicate()
```
lemaitre-aneo: Why is this test Postgres only?

Adaptors/PostgresSQL/src/ResultWatcher.cs
```diff
@@ -0,0 +1,205 @@
...
+                                                      return null;
+                                                    });
+  }
+
+  /// <inheritdoc />
+  public void Dispose()
+  {
+    insertBroadcaster_.Dispose();
+    updateBroadcaster_.Dispose();
+    GC.SuppressFinalize(this);
```
lemaitre-aneo:
```suggestion
```

lemaitre-aneo (dismissed):

lemaitre-aneo (approved):

PR #1271, opened 2026-07-28
fix: pageSize<=0 crash in MongoDB ListApplicationsAsync, dedup in Memory AddTaskDependencies

aneojgurhem:
# Motivation

While working on the Postgres adapter, two regression tests already existed for Postgres-specific fixes: a guard against negative `pageSize` in `ListApplicationsAsync`, and deduplication in `AddTaskDependencies`. Checking whether Memory and MongoDB satisfied the same behavioral contracts revealed that neither did — each had its own latent bug.

# Description

- **MongoDB `TaskTable.ListApplicationsAsync`**: a negative or zero `pageSize` was passed straight into `.Take(pageSize)`, which the MongoDB LINQ provider translates into a `$limit` pipeline stage. That stage's constructor rejects non-positive values and throws `ArgumentOutOfRangeException` instead of returning an empty result, unlike Postgres which already guards this case. Added the same `pageSize <= 0` guard, returning `(Enumerable.Empty<Application>(), totalCount)` without building the paginated query.
- **Memory `ResultTable.AddTaskDependencies`**: appended incoming task ids to `DependentTasks` unconditionally via `List<string>.AddRange`, so calling it twice with an overlapping task id produced a duplicate entry. MongoDB already dedups this via `$addToSet`. Fixed by excluding task ids already present in `DependentTasks` before appending (`taskIds.Except(result.DependentTasks)`).
- Promoted the two regression tests (`ListApplicationsAsyncShouldReturnEmptyForNegativePageSize`, `AddTaskDependenciesTwiceShouldNotDuplicate`) from the Postgres-specific test files into the shared `TaskTableTestBase`/`ResultTableTestBase`, so all three adapters (Memory, MongoDB, Postgres) are held to the same contracts going forward.

# Testing

- `dotnet test Adaptors/Memory/tests/` — 205/205 passed
- `dotnet test Adaptors/MongoDB/tests/` — 310/310 passed
- Both new shared tests verified to pass on Memory, MongoDB, and Postgres.

# Impact

- Behavioral fix only for two edge cases (negative `pageSize`, repeated `AddTaskDependencies` calls with overlapping ids); no change to normal usage paths.
- No new dependencies, no configuration or documentation changes.

# Additional Information

Extracted from work on the `jg/postgressql` branch, where Postgres already handled both cases correctly — this PR brings Memory and MongoDB in line with the same contracts.

### DIFF
--- a/Adaptors/Memory/src/ResultTable.cs
+++ b/Adaptors/Memory/src/ResultTable.cs
@@ -231,7 +231,8 @@ public class ResultTable : IResultTable
         throw new ResultNotFoundException($"Key '{resultId}' not found");
       }
 
-      result.DependentTasks.AddRange(taskIds);
+      var newTaskIds = taskIds.Except(result.DependentTasks);
+      result.DependentTasks.AddRange(newTaskIds);
     }
 
     return Task.CompletedTask;
--- a/Adaptors/MongoDB/src/TaskTable.cs
+++ b/Adaptors/MongoDB/src/TaskTable.cs
@@ -365,12 +365,19 @@ public class TaskTable : BaseTable<TaskData, TaskDataModelMapping>, ITaskTable
                                                               ascOrder)
                                       : queryable;
 
-    var taskResult = paged.Skip(page * pageSize)
-                          .Take(pageSize)
-                          .ToListAsync(cancellationToken);
+    var totalCount = paged.CountAsync(cancellationToken);
 
-    return (await taskResult.ConfigureAwait(false), await paged.CountAsync(cancellationToken)
-                                                               .ConfigureAwait(false));
+    if (pageSize <= 0)
+    {
+      return (Enumerable.Empty<Application>(), await totalCount.ConfigureAwait(false));
+    }
+
+    var taskResult = await paged.Skip(page * pageSize)
+                                .Take(pageSize)
+                                .ToListAsync(cancellationToken)
+                                .ConfigureAwait(false);
+
+    return (taskResult, await totalCount.ConfigureAwait(false));
   }
 
   /// <inheritdoc />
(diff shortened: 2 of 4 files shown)


Adaptors/Memory/src/ResultTable.cs
```diff
@@ -231,7 +231,9 @@ public Task AddTaskDependencies(IDictionary<string, ICollection<string>> depende
         throw new ResultNotFoundException($"Key '{resultId}' not found");
       }
 
-      result.DependentTasks.AddRange(taskIds);
+      var newTaskIds = taskIds.Except(result.DependentTasks)
+                              .ToList();
+      result.DependentTasks.AddRange(newTaskIds);
```
lemaitre-aneo: Is ToList actually required? I thought AddRange would take an IEnumerable directly.

Adaptors/MongoDB/src/TaskTable.cs
```diff
@@ -365,12 +365,20 @@ async Task<long> ITaskTable.BulkUpdateTasks(IEnumerable<(Expression<Func<TaskDat
...
+    {
+      return (Enumerable.Empty<Application>(), totalCount);
+    }
+
+    var taskResult = await paged.Skip(page * pageSize)
+                                .Take(pageSize)
+                                .ToListAsync(cancellationToken)
+                                .ConfigureAwait(false);
+
+    return (taskResult, totalCount);
```
lemaitre-aneo: This serializes both requests while they were executed in parallel before. I am not sure if that has an actual impact, though.

lemaitre-aneo (dismissed):

lemaitre-aneo (approved):

PR #1273, opened 2026-07-29
perf(postgres): project only selector-needed columns in reads

aneojgurhem:
# Motivation

`TaskTable.ReadTaskAsync<T>`/`ListTasksAsync<T>` and `ResultTable.GetResults<T>` in the Postgres adapter always executed `SELECT * FROM ...`, fetching every column regardless of what the caller's projection selector actually needed - and unconditionally ran a second join-table query to populate `TaskData.RemainingDataDependencies`/`Result.DependentTasks`, even when the caller's selector never read those fields.

# Description

- Added `Adaptors/PostgresSQL/src/Common/SelectorColumns.cs`: an expression-tree visitor that inspects a projection selector (`Expression<Func<TSource, TResult>>`) and determines the minimal set of SQL columns it needs, handling leaf members, compound members (e.g. `Options`/`Output`, expanded to their sub-columns), `.Count` accessors on array fields, and falling back to the full row for identity selectors or unrecognized shapes.
- `RowMapper.MapToTaskData`/`MapToResult` gained a partial-column overload that defaults any field whose column wasn't selected, so callers passing a narrowed column set still get a valid object back.
- `TaskTable.ReadTaskAsync<T>` and `ListTasksAsync<T>`, and `ResultTable.GetResults<T>`, now use `SelectorColumns` to build a narrower `SELECT`, and only run the separate join-table query (`task_remaining_dependencies` / `result_dependent_tasks`) when the selector actually reads the corresponding field.
- Important asymmetry, verified by tracing every real caller of both methods: `TaskData.RemainingDataDependencies` has no caller that needs it via an identity selector (by the time `EndTaskAsync`/`RetryTask` reads it, the task has already left `Creating`/`Pending`, so it's guaranteed empty anyway) - so the identity/fallback case skips loading it. `Result.DependentTasks` is different: `TaskLifeCycleHelper.ResolveDependencies` reads it off results obtained via the identity selector and relies on it being genuinely populated to resolve task readiness, so the identity/fallback case for `Result` still loads it.
- `TaskTable.ListTasksAsync<T>` was prototyped with the same treatment but reverted: it would have required abandoning MongoDB's efficient `.Project(selector)` server-side projection for zero measured benefit, since the real production selectors already request nearly every column anyway.

# Testing

- Added `ReadTaskAsyncShouldReturnRemainingDataDependenciesWhenProjected` (shared, all 3 adapters) and `ReadTaskAsyncShouldNotReturnRemainingDataDependencies` (Postgres-only, since Memory/MongoDB have no join query to avoid and always return the field).
- Added `GetResultsShouldReturnDependentTasksWithIdentitySelector` (shared) as a regression guard for the critical `TaskLifeCycleHelper.ResolveDependencies` path.
- Full adapter test suites pass: Postgres 425/425, Memory 222/222, MongoDB 327/327.

# Impact

- Reduces query size/columns fetched for narrow projections in `ReadTaskAsync`/`GetResults<T>`; many real `GetResults<T>` callers use narrow single/few-field selectors, so this is a genuine efficiency win there. `ListTasksAsync<T>`'s real callers (via `TaskDataMask`) already request nearly every column, so the win there is smaller/moot in practice but the mechanism is in place for future narrower selectors.
- No behavior change for any currently-known production caller (verified by tracing all callers of the affected methods).
- No new dependencies, no configuration changes.


Adaptors/PostgresSQL/src/Common/SelectorColumns.cs
```diff
@@ -0,0 +1,241 @@
...
+  // selector doesn't ask for it explicitly, identity included.
+  //
+  // Result.DependentTasks: TaskLifeCycleHelper.ResolveDependencies reads it off results obtained
+  // via the identity selector and relies on it being genuinely populated to resolve task
+  // readiness - so, unlike TaskData, the identity/fallback case must still load it.
+  private static readonly IReadOnlyDictionary<Type, (string FieldName, string IdFieldPath, bool LoadOnFallback)> SeparatelyStoredFields =
+    new Dictionary<Type, (string, string, bool)>
+    {
+      [typeof(TaskData)] = ("RemainingDataDependencies", "TaskId", false),
+      [typeof(Result)] = ("DependentTasks", "ResultId", true),
```
aneojgurhem: can we use nameof instead ?

Adaptors/PostgresSQL/src/Common/RowMapper.cs
```diff
@@ -37,64 +37,109 @@ public static class RowMapper
   /// <param name="reader">The data reader</param>
   /// <returns>The TaskData record</returns>
   public static TaskData MapToTaskData(NpgsqlDataReader reader)
+    => MapToTaskData(reader,
+                     null);
```
lemaitre-aneo: This overload could be omitted if the other variant had a null default value.

Adaptors/PostgresSQL/src/Common/SelectorColumns.cs
```diff
@@ -0,0 +1,241 @@
...
+  //
+  // TaskData.RemainingDataDependencies: the only real consumer (TaskLifeCycleHelper's
+  // EndTaskAsync/RetryTask path) reads it off a task that has already left Creating/Pending, at
+  // which point it is guaranteed empty by the state machine - so it is safe to skip whenever the
+  // selector doesn't ask for it explicitly, identity included.
+  //
+  // Result.DependentTasks: TaskLifeCycleHelper.ResolveDependencies reads it off results obtained
+  // via the identity selector and relies on it being genuinely populated to resolve task
+  // readiness - so, unlike TaskData, the identity/fallback case must still load it.
+  private static readonly IReadOnlyDictionary<Type, (string FieldName, string IdFieldPath, bool LoadOnFallback)> SeparatelyStoredFields =
```
lemaitre-aneo: Create a local record struct for the Value type

Adaptors/PostgresSQL/tests/TaskTableTests.cs
```diff
@@ -43,4 +46,24 @@ public override void GetTaskTableInstance()
     TaskTable = provider.GetRequiredService<ITaskTable>();
     RunTests  = true;
   }
+
+  [Test]
+  public async Task ReadTaskAsyncShouldNotReturnRemainingDataDependencies()
```
lemaitre-aneo, line 51: Should not be PostgreSQL only

Adaptors/PostgresSQL/src/Common/RowMapper.cs
aneojgurhem: Done — both overloads are gone, `columns` now defaults to `null` on `MapToTaskData` and `MapToResult`.

Note for anyone touching this again: `RowMapperTests.ColumnCases` locates each mapping method by searching the source for `" <Method>(NpgsqlDataReader reader)"`, so removing the single-argument signatures broke the marker. It now matches on `" <Method>(NpgsqlDataReader"` — parameter type only — which tolerates both the extra optional argument and the alignment padding the formatter adds.

Adaptors/PostgresSQL/src/Common/SelectorColumns.cs
aneojgurhem: Done — added a `private readonly record struct SeparatelyStoredField(string FieldName, string IdFieldPath, bool LoadOnFallback)` and the dictionary is now `IReadOnlyDictionary<Type, SeparatelyStoredField>`. Reads a lot better at the use sites too (`separateField.LoadOnFallback` instead of a positional tuple element).

Adaptors/PostgresSQL/tests/TaskTableTests.cs
aneojgurhem: I'd like to keep this one PostgreSQL-only, because it asserts a PostgreSQL-specific optimisation rather than a contract of `ITaskTable`.

`RemainingDataDependencies` is the only `TaskData` field that does not live in the `tasks` table — it is stored in the `task_remaining_dependencies` join table, so populating it costs an extra query. This PR skips that query unless the selector actually asks for the field, and this test is what pins that down. Memory and MongoDB have the field sitting in the stored object already, so there is nothing for them to skip and no equivalent behaviour to assert.

Promoting the test to `TaskTableTestBase` would make it fail on both of those adapters (`TaskCreatingId` has `dependency1` as a data dependency, so they return it populated). Getting them to pass would mean stripping the field client-side, which for MongoDB means giving up the server-side `.Project(selector)` on `ReadTaskAsync` — a real cost for no caller's benefit, and something we deliberately decided against earlier.

The cross-adapter direction that *is* a genuine contract — project the field explicitly and it must come back populated — is covered in the shared base by `ReadTaskAsyncShouldReturnRemainingDataDependenciesWhenProjected`, alongside the equivalents for `ListTasksAsync` and `UpdateOneTask`.

Happy to revisit if you think the population of that field should be a uniform guarantee across adapters — that is a bigger change than this PR, though, and it would cost MongoDB its projection pushdown.

Adaptors/PostgresSQL/src/TaskTable.cs
```diff
@@ -352,10 +363,26 @@ await cmd.ExecuteNonQueryAsync(cancellationToken)
...
+      var depsByTaskId = await LoadRemainingDependenciesBatch(connection,
+                                                              rawTasks,
+                                                              cancellationToken)
+                           .ConfigureAwait(false);
+
+      rawTasks = rawTasks.Select(taskData => taskData with
+                                             {
+                                               RemainingDataDependencies = depsByTaskId[taskData.TaskId],
+                                             })
+                         .ToList();
```
lemaitre-aneo: I think you can change in place instead of materializing a new copy.

lemaitre-aneo (dismissed):

Adaptors/PostgresSQL/src/TaskTable.cs
```diff
@@ -376,11 +376,13 @@ await cmd.ExecuteNonQueryAsync(cancellationToken)
...
-                                               RemainingDataDependencies = depsByTaskId[taskData.TaskId],
-                                             })
-                         .ToList();
+      for (var i = 0; i < rawTasks.Count; i++)
+      {
+        rawTasks[i] = rawTasks[i] with
+                      {
+                        RemainingDataDependencies = depsByTaskId[rawTasks[i].TaskId],
+                      };
+      }
```
lemaitre-aneo, line 385:
```suggestion
      foreach (var rawTask in rawTasks)
      {
        rawTask.RemainingDataDependencies = depsByTaskId[rawTask.TaskId];
      }
```

aneojgurhem: taskdata is a record

lemaitre-aneo (approved):

PR #1276, opened 2026-07-30
perf(postgres): run List* count and page queries in parallel

aneojgurhem:
# Motivation

Every paginated `List*` method in the PostgreSQL adapter issues two independent queries — a `COUNT(*)` over the filter, and a `LIMIT`/`OFFSET` page — and ran them one after the other on a single connection. Since neither depends on the other, the caller was paying `count + page` when it only needs to pay `max(count, page)`.

The `COUNT(*)` is usually the more expensive of the two: it has to visit every row matching the filter, whereas the page query is bounded by `pageSize`. So the wasted time is roughly the whole duration of the page query, on every list call, on paths that back the public gRPC list endpoints.

Those five methods were also near-identical copies of one another, so the parallelisation is factored into one shared helper rather than pasted five times.

# Description

New `Common/PagedQuery.cs` owns both the concurrency and the query generation, and exposes two entry points:

| Overload | Used by | Generates |
|---|---|---|
| `ExecuteAsync<TEntity, TRow>` | the four regular list methods | the filter/order translation, `SELECT COUNT(*) FROM {table} WHERE …`, and `SELECT {selectList} FROM {table} WHERE … ORDER BY … LIMIT @limit OFFSET @offset` |
| `ExecuteAsync<TRow>` | `ListApplicationsAsync` | nothing — the caller supplies both queries as `Query` values |

Affected methods: `ListSessionsAsync`, `ListResultsAsync`, `ListPartitionsAsync`, `ListTasksAsync<T>`, `ListApplicationsAsync`.

Six things that were duplicated across the four regular callers now live in one place: the count SQL, the page SQL, `ExpressionToSql.Translate`, `TranslateOrderBy`, the order-parameter merge, and the `ascOrder ? "ASC" : "DESC"` ternary. `ListSessionsAsync` is a single statement. The SQL produced is unchanged.

Points worth a reviewer's attention:

**Each query takes its own pooled connection.** A single `NpgsqlConnection` can only run one command at a time. `NpgsqlConnectionProvider.GetConnectionAsync` delegates to `NpgsqlDataSource.OpenConnectionAsync`, which hands out a distinct pooled connection per call.

**Snapshot consistency is unchanged.** The two queries now run on separate connections rather than sequentially on one, but they were never in a shared transaction — under `READ COMMITTED` each statement already took its own snapshot. A count that disagrees slightly with the page was possible before and remains possible now.

**`pageSize <= 0` still short-circuits.** The count is launched first; when the page would be empty the helper returns without launching the page query.

**`Task.WhenAll` rather than two sequential awaits**, so if one query faults the other is still observed instead of surfacing later as an unobserved task exception.

**The two queries do not share a parameter dictionary.** Only the page references the `ORDER BY` parameters, so the count is given the filter's set alone. Reversing that merge is safe because the namespaces are disjoint — `ExpressionToSql.Translate` names filter parameters `@p<n>`, `TranslateOrderBy` names order keys `@orderkey<n>`.

`ListApplicationsAsync` collects its order parameters into their own dictionary rather than adding them to the filter's. That removes the defensive copy the count previously needed: nothing is mutated after either `Query` is built, so there is no dictionary being written to while the other query reads it — a structural guarantee rather than a guarded one.

The helper returns `long` for the count; the three methods whose interfaces return `int` convert with `Convert.ToInt32`, preserving the existing overflow-throwing behaviour rather than silently wrapping.

# Testing

| Suite | Result |
|---|---|
| `Adaptors/PostgresSQL/tests` | **441/441** |
| `Adaptors/MongoDB/tests` | **330/330** |
| `Adaptors/Memory/tests` | **225/225** |

Adds one test: **`ListSessionsAsyncOrderByTaskOptionShouldSucceed`**, in the shared `SessionTableTestBase`. It orders on a task option (`data.Options.Options["sortkey"]`) rather than on a column, asserting both the count and the resulting order.

That shape matters. `TranslateOrderBy` only produces a parameter of its own for the dictionary-indexer form; for an ordinary column it returns an empty dictionary. So ordering by a task option is the only way a test can notice a paginated read handing the wrong parameters to its count or to its page — and before this, the only such coverage anywhere was one `ListTasksAsync` test for quoted keys. The count/page split described above was otherwise unverified, and the old code worked because the surplus parameters were harmless rather than because they were right.

It lives in the shared base deliberately, so it holds for Memory and MongoDB too; all three adapters gained exactly one test, which confirms it ran on each rather than being skipped.

# Impact

No API, schema, configuration, or dependency changes. Behaviour is intended to be identical; only the execution shape changes.

Single-call latency, medians against PostgreSQL 18.4, filter matching about one sixth of the table:

| Table rows | pageSize | sequential | parallel | |
|---|---|---|---|---|
| 50 000 | 20 | 7.49 ms | 4.05 ms | 45.9% faster |
| 50 000 | 500 | 8.21 ms | 5.16 ms | 37.1% faster |
| 400 000 | 20 | 54.98 ms | 35.74 ms | 35.0% faster |
| 400 000 | 500 | 55.45 ms | 36.47 ms | 34.2% faster |

The trade-off is that a list call holds two pooled connections instead of one for its duration, so it was worth checking under pool pressure: **64 concurrent list calls** — 128 connections wanted against the default `MaxPoolSize` of 100 — went from **1192.9 ms to 1073.0 ms**, still around 10% faster, with no connection timeouts.

Deadlock is not possible: the count never waits on the page, so even a fully saturated pool always drains. The run above exercised exactly that oversubscribed case.

Caveat on the figures: measured over WSL2-to-Docker loopback, so absolute numbers are not production figures. The effect comes from overlapping two server-side queries rather than from network latency, so the direction holds, but the percentages will vary with the relative cost of the count and page for a given filter.

# Additional Information

Rebased onto `main` after #1273 merged. That rebase surfaced one thing git could not: `RowMapper.MapToResult` became an instance method in #1273, and this branch passed it as a static method group in `ListResultsAsync` — a compile error rather than a conflict, so `git merge-tree` reported a clean automatic merge for that file. It now uses `RowMapper.FullRow.MapToResult`.

`ListTasksAsync<T>` carries both features: `RowMapper.For(selector)` supplies the projected column list and mapper from #1273, while the count and page run concurrently. Its dependency batch-load opens its own connection, since `PagedQuery` owns the ones it queries on.

Formatting was applied from the `Code Formatting` job's patch artifact rather than run locally.

# Checklist

### DIFF
--- a/Adaptors/PostgresSQL/src/Common/PagedQuery.cs
+++ b/Adaptors/PostgresSQL/src/Common/PagedQuery.cs
@@ -26,13 +26,12 @@ using Npgsql;
 namespace ArmoniK.Core.Adapters.PostgresSQL.Common;
 
 /// <summary>
-///   The page half of a paginated query: its SQL, which must end with <c>LIMIT @limit OFFSET @offset</c>,
-///   and the parameters its WHERE and ORDER BY clauses reference.
+///   A query and the parameters it references.
 /// </summary>
-/// <param name="Sql">The page query</param>
-/// <param name="Parameters">Parameters referenced by <paramref name="Sql" />, excluding limit and offset</param>
-public readonly record struct PageQuery(string                      Sql,
-                                        Dictionary<string, object?> Parameters);
+/// <param name="Sql">The query</param>
+/// <param name="Parameters">Parameters referenced by <paramref name="Sql" /></param>
+public readonly record struct Query(string                      Sql,
+                                    Dictionary<string, object?> Parameters);
 
 /// <summary>
 ///   Runs the two queries behind a paginated list: the total number of matches, and one page of rows.
@@ -47,7 +46,7 @@ public static class PagedQuery
   /// </summary>
   /// <remarks>
   ///   Queries that don't have that shape - one counting distinct combinations of columns, say, or one
-  ///   ordering on several fields - build their own SQL and call the overload taking it directly.
+  ///   ordering on several fields - build their own and call the overload taking them directly.
   /// </remarks>
   /// <typeparam name="TEntity">Entity type the filter and order field apply to</typeparam>
   /// <typeparam name="TRow">Type each row maps to</typeparam>
@@ -79,9 +78,13 @@ public static class PagedQuery
   {
     var (whereSql, parameters)     = ExpressionToSql<TEntity>.Translate(filter);
     var (orderColumn, orderParams) = ExpressionToSql<TEntity>.TranslateOrderBy(orderField);
-    foreach (var (key, value) in orderParams)
+
+    // Only the page references the ORDER BY parameters, so they are kept out of the count's set rather
+    // than merged into it. The two namespaces are disjoint - filters are named @p<n> and order keys
+    // @orderkey<n> - so nothing is overwritten either way round.
+    foreach (var (key, value) in parameters)
     {
-      parameters[key] = value;
+      orderParams[key] = value;
     }
 
     var orderDir = ascOrder
@@ -89,12 +92,12 @@ public static class PagedQuery
                      : "DESC";
 
     return await ExecuteAsync(connectionProvider,
-                              $"SELECT COUNT(*) FROM {table} WHERE {whereSql}",
-                              parameters,
+                              new Query($"SELECT COUNT(*) FROM {table} WHERE {whereSql}",
+                                        parameters),
+                              new Query($"SELECT {selectList} FROM {table} WHERE {whereSql} ORDER BY {orderColumn} {orderDir} LIMIT @limit OFFSET @offset",
+                                        orderParams),
                               page,
                               pageSize,
-                              () => new PageQuery($"SELECT {selectList} FROM {table} WHERE {whereSql} ORDER BY {orderColumn} {orderDir} LIMIT @limit OFFSET @offset",
-                                                  parameters),
                               mapRow,
                               cancellationToken)
              .ConfigureAwait(false);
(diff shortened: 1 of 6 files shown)


lemaitre-aneo (commented):
As the queries are type agnostics, it should be easy to factor this logic out and avoid code duplication.

Adaptors/PostgresSQL/src/Common/PagedQuery.cs
```diff
@@ -0,0 +1,205 @@
...
+  /// <param name="pageQuery">
+  ///   Builds the page query. Only called when <paramref name="pageSize" /> is positive, so any cost of
+  ///   assembling it (translating ORDER BY fields, say) is skipped when only the count was asked for.
+  /// </param>
+  /// <param name="mapRow">Maps the reader's current row to <typeparamref name="TRow" /></param>
+  /// <param name="cancellationToken">Token used to cancel the execution of the method</param>
+  /// <returns>The page's rows, and the total number of matches irrespective of paging</returns>
+  public static async Task<(List<TRow> items, long totalCount)> ExecuteAsync<TRow>(NpgsqlConnectionProvider     connectionProvider,
+                                                                                   string                       countSql,
+                                                                                   Dictionary<string, object?>  countParameters,
```
lemaitre-aneo: Could be a PageQuery parameter

Adaptors/PostgresSQL/src/Common/PagedQuery.cs
```diff
@@ -0,0 +1,205 @@
...
+                                                                                   int                          page,
+                                                                                   int                          pageSize,
+                                                                                   Func<PageQuery>              pageQuery,
```
lemaitre-aneo: There is no need for a closure here. just the plain value should be enough

Adaptors/PostgresSQL/src/Common/PagedQuery.cs
```diff
@@ -0,0 +1,205 @@
...
+namespace ArmoniK.Core.Adapters.PostgresSQL.Common;
+
+/// <summary>
+///   The page half of a paginated query: its SQL, which must end with <c>LIMIT @limit OFFSET @offset</c>,
+///   and the parameters its WHERE and ORDER BY clauses reference.
+/// </summary>
+/// <param name="Sql">The page query</param>
+/// <param name="Parameters">Parameters referenced by <paramref name="Sql" />, excluding limit and offset</param>
+public readonly record struct PageQuery(string                      Sql,
+                                        Dictionary<string, object?> Parameters);
```
lemaitre-aneo: There is no paging related field here.
```suggestion
public readonly record struct Query(string                      Sql,
                                    Dictionary<string, object?> Parameters);
```

Adaptors/PostgresSQL/src/Common/PagedQuery.cs
```diff
@@ -0,0 +1,196 @@
...
+                                                                                            Func<NpgsqlDataReader, TRow>       mapRow,
+                                                                                            CancellationToken                  cancellationToken,
+                                                                                            string                             selectList = "*")
+  {
+    var (whereSql, parameters)     = ExpressionToSql<TEntity>.Translate(filter);
+    var (orderColumn, orderParams) = ExpressionToSql<TEntity>.TranslateOrderBy(orderField);
+    foreach (var (key, value) in orderParams)
+    {
+      parameters[key] = value;
+    }
```
lemaitre-aneo:
```suggestion
    foreach (var (key, value) in parameters)
    {
      orderParams[key] = value;
    }
```

Adaptors/PostgresSQL/src/Common/PagedQuery.cs
```diff
@@ -0,0 +1,196 @@
...
+
+    var orderDir = ascOrder
+                     ? "ASC"
+                     : "DESC";
+
+    return await ExecuteAsync(connectionProvider,
+                              new Query($"SELECT COUNT(*) FROM {table} WHERE {whereSql}",
+                                        parameters),
+                              new Query($"SELECT {selectList} FROM {table} WHERE {whereSql} ORDER BY {orderColumn} {orderDir} LIMIT @limit OFFSET @offset",
+                                        parameters),
```
lemaitre-aneo:
```suggestion
                              new Query($"SELECT {selectList} FROM {table} WHERE {whereSql} ORDER BY {orderColumn} {orderDir} LIMIT @limit OFFSET @offset",
                                        orderParams),
```

lemaitre-aneo (approved):

PR #1279, opened 2026-07-31
test(pollster): trigger the pollster tests on events instead of delays

aneojgurhem:
# Motivation

`testsWinOnly (Common/tests)` and `tests (Common/tests, ubuntu-latest)` fail intermittently, reddening unrelated pull requests — most recently #1277, where the single failure was:

```
Failed RunThenCancelPollster
  Expected: No Exception to be thrown
  But was:  <System.Threading.Tasks.TaskCanceledException: A task was canceled.
   at TestPollsterProvider.<<StopApplicationAfter>b__0>d.MoveNext() ... :line 246
```

The mechanism: `Pollster.MainLoop` runs until the application stops, so a test has to stop it before it can assert anything. `StopApplicationAfter` did that after a fixed delay, awaiting `ExceptionManager.EarlyCancellationToken`. But `MainLoop`'s `finally` calls `UnregisterAndStop`, which cancels that very token — so a pollster that ends on its own cancels the pending delay, and the task faults instead of firing. `RunThenCancelPollster` asked for a 105 µs delay against a timer granularity of roughly 15 ms, so it lost that race whenever the loop finished promptly. It always raced; it was merely usually lucky. It failed on **Linux** in #1277, so this was never a Windows artifact.

Rather than tolerate the exception, this replaces the delays with the events they were standing in for.

# Description

`StopApplicationAfter` is gone. In its place, on `TestPollsterProvider`:

- **`StopApplicationWhen(Task trigger)`** — awaits the trigger, then stops the application. Deliberately not cancellable, so a pollster that ends first cannot fault it.
- **`WaitForTaskStatus(taskId, statuses, timeout)`** — for what a test asserts about the database. On timeout it reports which status was actually reached.
- **`WaitForNoTaskProcessing(timeout)`** — draining `TaskProcessing` is not instantaneous and the loop ending does not guarantee it.

The worker fakes (`SimpleWorkerStreamHandler`, `WaitWorkerStreamHandler`, `ExceptionWorkerStreamHandler<T>`) expose a `Started` task, completed when the pollster hands them a task, using `RunContinuationsAsynchronously` so a test's continuation does not run inline on the pollster's thread mid-handover.

**Six of the seven tests** now trigger off an event or an observed state:

| Test | Was | Now |
|---|---|---|
| `RunThenCancelPollster` | 105 µs | `TaskCompletionSource` signalled from the queue mock |
| `ExecuteTaskShouldSucceed` | 1 s | `worker.Started`, then `WaitForTaskStatus(Completed)` after the loop returns |
| `ExecuteTaskTimeoutAcquire` | 2 s | `worker.Started`; the second task stays `Submitted` via the 10 s `TimeoutBeforeNextAcquisition` |
| `CancelLongTaskShouldSucceed` | 5 s + 200 ms + 200 ms | `worker.Started`, `WaitForTaskStatus(Cancelling/Cancelled)`, `WaitForNoTaskProcessing()`, and a `TaskCompletionSource` the test completes when done driving |
| `ExecuteTooManyErrorShouldFail` | 10 s guard | no stop scheduled at all |
| `UnavailableWorkerShouldFail` | 300 ms guard | no stop scheduled at all |

The last two are the neatest outcome: the pollster is *expected* to give up on its own, so their delays existed only to be outlived. They now assert the outcome directly, on `EarlyCancellationToken.AsTask()` throwing.

## A rule this exercise established

**Waiting for a final task status has to happen after the loop returns, not before stopping.** `MainLoop`'s `finally` closes the running-task queue without awaiting the post processor, so a task can reach its final state only as the application shuts down. An earlier revision of this PR triggered the stop *on* that state, which deadlocked `ExecuteTaskShouldSucceed` until its 10 s timeout on the Windows runner — green on Linux and locally, red on CI. Both affected tests now stop on an in-flight event and wait for the final state afterwards.

## One test is deliberately not converted

`ExecuteTaskThatExceedsGraceDelayShouldResubmit` **keeps both of its delays and keeps `[Retry(3)]`.** Its `AssertFailAfterError(5)` only passes if the pollster has already recorded an error, and how many errors it records is a function of how long it ran — so those delays feed the assertions rather than waiting for anything observable. Three attempts to convert it each cut the run short and left the error count too low. Converting it properly means changing what it asserts, which is a separate job.

It does still benefit: its stop goes through `StopApplicationWhen(Task.Delay(300, CancellationToken.None))`, so it no longer awaits the early cancellation token and cannot fault the way the original did.

## A latent bug fixed along the way

`RunThenCancelPollster`'s queue mock matched the partition `"partitionId"`, while `TestPollsterProvider` configures `"DefaultPartition"`. The setup therefore never matched, `PullMessagesAsync` returned null, and the pollster died on a `NullReferenceException` — that is the NRE visible in CI logs. The test had never once exercised the pull it claimed to; it passed only because its delay stopped the application regardless. Making the trigger honest is what exposed this.

`[Retry(3)]` is removed from the three other tests that carried it, since it was masking the race.

# Testing

No new tests — this repairs existing ones.

| Run | Result |
|---|---|
| Full `Common/tests`, run 1 | **7192 passed** (6m18s) |
| Full `Common/tests`, run 2 | **7192 passed** (7m43s) |

Two *full-suite* runs rather than repeated runs of the fixture alone, and that distinction matters. The defect that the earlier revision of this PR contained was only caught when the machine was loaded — the fixture in isolation passed three times in a row while a real problem was present. The full suite is what generates the contention, and run 2 at 7m43s is a slow, loaded pass.

Both runs report 2 failures, `Eviction` and `Eviction2`. Both live in an untracked local scratch file (`Common/tests/LolTests.cs`, unknown to git and so never compiled in CI). Everything tracked passes, with no failure in this fixture.

Honest limitation: the `ExecuteTaskShouldSucceed` timeout only ever appeared on the Windows runner and cannot be reproduced locally, so CI's `testsWinOnly (Common/tests)` is the only real check on that part. Removing `[Retry(3)]` is deliberate on that point — if a race survives, CI will now say so instead of retrying it away.

# Impact

Test-only; no product code touched. Nothing ships.

The fixture gets substantially faster, and — more to the point — a failure in it now means something. The helpers are reusable for the same pattern elsewhere: `TaskHandlerTest` has comparable `Task.Delay`-then-assert sequences that were left alone here.

# Additional Information

Deliberately out of scope: the same stale `"partitionId"` literal appears in `ExecuteTooManyErrorShouldFail`'s mock, so its "too many errors" also arises from a null-returning mock rather than the scenario it names. It passes, and correcting it would change what that test exercises, so it is worth its own change.

Formatting was applied from the `Code Formatting` job's patch artifact rather than run locally.

### DIFF
--- a/Common/tests/Helpers/ExceptionWorkerStreamHandler.cs
+++ b/Common/tests/Helpers/ExceptionWorkerStreamHandler.cs
@@ -37,8 +37,8 @@ public sealed class ExceptionWorkerStreamHandler<T> : IWorkerStreamHandler
   private readonly int  delay_;
   private readonly bool healthy_;
 
-  // RunContinuationsAsynchronously so whatever a test hangs off Started does not run inline on the
-  // pollster's thread, in the middle of it handing the task over.
+  // Continuations run on the thread pool: a test awaiting Started must not resume on the pollster's own
+  // thread while it is handing the task over, or test and pollster serialise on that thread.
   private readonly TaskCompletionSource started_ = new(TaskCreationOptions.RunContinuationsAsynchronously);
 
   public ExceptionWorkerStreamHandler(int  delay,
--- a/Common/tests/Helpers/ExceptionWorkerStreamHandler.cs
+++ b/Common/tests/Helpers/ExceptionWorkerStreamHandler.cs
@@ -37,6 +37,10 @@ public sealed class ExceptionWorkerStreamHandler<T> : IWorkerStreamHandler
   private readonly int  delay_;
   private readonly bool healthy_;
 
+  // RunContinuationsAsynchronously so whatever a test hangs off Started does not run inline on the
+  // pollster's thread, in the middle of it handing the task over.
+  private readonly TaskCompletionSource started_ = new(TaskCreationOptions.RunContinuationsAsynchronously);
+
   public ExceptionWorkerStreamHandler(int  delay,
                                       bool acceptCancellation = true,
                                       bool healthy            = true)
@@ -45,6 +49,12 @@ public sealed class ExceptionWorkerStreamHandler<T> : IWorkerStreamHandler
     acceptCancellation_ = acceptCancellation;
     healthy_            = healthy;
   }
+
+  /// <summary>
+  ///   Completes the first time the pollster hands a task to this worker, before it starts waiting.
+  /// </summary>
+  public Task Started
+    => started_.Task;
 
   public Task<HealthCheckResult> Check(HealthCheckTag tag)
     => Task.FromResult(healthy_
@@ -64,6 +74,8 @@ public sealed class ExceptionWorkerStreamHandler<T> : IWorkerStreamHandler
                                                 Configuration     configuration,
                                                 CancellationToken cancellationToken)
   {
+    started_.TrySetResult();
+
     await Task.Delay(TimeSpan.FromMilliseconds(delay_),
                      acceptCancellation_
                        ? cancellationToken
--- a/Common/tests/Helpers/SimpleWorkerStreamHandler.cs
+++ b/Common/tests/Helpers/SimpleWorkerStreamHandler.cs
@@ -32,8 +32,8 @@ namespace ArmoniK.Core.Common.Tests.Helpers;
 
 public class SimpleWorkerStreamHandler : IWorkerStreamHandler
 {
-  // RunContinuationsAsynchronously so whatever a test hangs off Started does not run inline on the
-  // pollster's thread, in the middle of it handing the task over.
+  // Continuations run on the thread pool: a test awaiting Started must not resume on the pollster's own
+  // thread while it is handing the task over, or test and pollster serialise on that thread.
   private readonly TaskCompletionSource started_ = new(TaskCreationOptions.RunContinuationsAsynchronously);
 
   public Output Output = new(OutputStatus.Success,
--- a/Common/tests/Helpers/SimpleWorkerStreamHandler.cs
+++ b/Common/tests/Helpers/SimpleWorkerStreamHandler.cs
@@ -32,8 +32,18 @@ namespace ArmoniK.Core.Common.Tests.Helpers;
 
 public class SimpleWorkerStreamHandler : IWorkerStreamHandler
 {
+  // RunContinuationsAsynchronously so whatever a test hangs off Started does not run inline on the
+  // pollster's thread, in the middle of it handing the task over.
+  private readonly TaskCompletionSource started_ = new(TaskCreationOptions.RunContinuationsAsynchronously);
+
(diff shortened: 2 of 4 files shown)


Common/tests/Helpers/ExceptionWorkerStreamHandler.cs
```diff
@@ -37,6 +37,10 @@ public sealed class ExceptionWorkerStreamHandler<T> : IWorkerStreamHandler
   private readonly int  delay_;
   private readonly bool healthy_;
 
+  // RunContinuationsAsynchronously so whatever a test hangs off Started does not run inline on the
+  // pollster's thread, in the middle of it handing the task over.
```
lemaitre-aneo: The comment is confusing.

Common/tests/Helpers/SimpleWorkerStreamHandler.cs
```diff
@@ -32,9 +32,19 @@ namespace ArmoniK.Core.Common.Tests.Helpers;
 
 public class SimpleWorkerStreamHandler : IWorkerStreamHandler
 {
+  // RunContinuationsAsynchronously so whatever a test hangs off Started does not run inline on the
+  // pollster's thread, in the middle of it handing the task over.
```
lemaitre-aneo: Confusing

Common/tests/Helpers/TestPollsterProvider.cs
```diff
@@ -240,15 +258,87 @@ public void Dispose()
...
+  ///   actually waiting for and neither problem exists - the trigger is deliberately not cancellable, so
+  ///   a pollster that ends first cannot fault this task either.
+  /// </remarks>
+  /// <param name="trigger">Task whose completion means the test has seen what it was waiting for</param>
+  public async Task StopApplicationWhen(Task trigger)
+  {
+    await trigger.ConfigureAwait(false);
+
+    Lifetime.StopApplication();
+  }
```
lemaitre-aneo: Shouldn't it be wrapped in a Task.Run()?

Common/tests/Pollster/PollsterTest.cs
```diff
@@ -631,6 +671,8 @@ await testServiceProvider.Pollster.Init(CancellationToken.None)
 
   [Test]
   [Timeout(10000)]
+  // Kept, unlike on the rest of the fixture: this test still depends on how much the pollster got done
+  // in a fixed time, for the reason given at its stop below.
   [Retry(3)]
   public async Task ExecuteTaskThatExceedsGraceDelayShouldResubmit([Values] bool healthy)
```
lemaitre-aneo: I'm somehow convinced that you can still reimplement it event-based.

lemaitre-aneo (approved):

PR #1296, opened 2026-09-17
chore(renovate): split nuget updates and add monthly schedule

aneojgurhem:
# Motivation

The `nuget packages` Renovate group is the only one that does not merge. Its history:

| PR | Created | Merged | Blocked |
|---|---|---|---|
| #395 | 2023-06-08 | 2024-03-01 | ~9 months |
| #699 | 2024-04-23 | 2025-12-11 | ~19 months |
| #1159 | 2026-02-17 | *open* | 7 months |
| #402 (major) | 2023-06-09 | *open* | 3+ years |

The terraform, github-actions, `ArmoniK.Api` and `ArmoniK.Utils` groups merge within days. The difference is batch size, not package quality: the nuget group bundles ~40 packages into one all-or-nothing PR, and Renovate rebases it continuously, so it only ever grows.

On #1159, the entire integration matrix is green and all 373 unit-test failures come from exactly **2 packages out of ~40**:

- **NUnit 4.4 -> 4.6** rejects `[Timeout]` on non-.NET-Framework target frameworks — 106 failures in `Common/tests`.
- **MongoDB.Driver 3.9.0** raised its SharpCompress floor from 0.30.1 to 0.48.1, breaking `MysticMind.PostgresEmbed` 4.0.0 which needs 0.34.2 — 267 failures in `Adaptors/PostgresSQL/tests`.

So 38 harmless updates have been held hostage for 7 months. A 40-package PR also forces an all-or-nothing trust decision, which is a large part of why it never gets reviewed.

# Description

Only `renovate.json` changes. Three things:

**1. Split the nuget group by release train.** Packages that ship together stay together (one changelog to read); everything else gets its own PR.

- Grouped: `.NET runtime` (`Microsoft.Extensions.*`, `Microsoft.Bcl.*`, `Microsoft.AspNetCore.*`), `gRPC`, `AWS SDK`, `Google Cloud`, `OpenTelemetry`, `NUnit`, `NATS`, `Serilog`.
- One PR each: `MongoDB.*`, `Npgsql`, `StackExchange.Redis`, `AMQPNetLite.Core`, `dbup-postgresql`, `Microsoft.NET.Test.Sdk`, `coverlet.*`, `Moq`, `EphemeralMongo*`, `MysticMind.PostgresEmbed` — the storage/queue drivers and test tooling, where a behaviour change is costly and the changelog is worth reading before merging.
- The existing `nuget packages` rule stays as a catch-all safety net for anything new.

**2. Monthly schedule on third-party nuget** (`on the first day of the month`). Over the 7 months #1159 was blocked, AWSSDK.S3 published **177 releases** (AWSSDK.Core 143, AWSSDK.SQS 126), against MongoDB.Driver 5 and Npgsql 7 — merging AWS today means seeing it again tomorrow. A monthly window makes that one PR per month.

It also freezes rebasing between windows, so a PR under review stops moving underneath the reviewer: #1159 was created 2026-02-17 and was still being rebased on 2026-09-17.

`ArmoniK.Api`, `ArmoniK.Utils` and `Htc.Mock` keep their current immediate cadence — the schedule sits on the catch-all rule, which already excludes those prefixes. Security fixes are unaffected: `vulnerabilityAlerts` ignores `schedule` by default.

**3. Allow prereleases for `OpenTelemetry.Exporter.Prometheus.*`.** This package has been beta since 2021 and sits at `1.15.3-beta.1` while the stable OTel packages move to `1.18.0`. With `ignoreUnstable` at its default, Renovate only accepts prerelease bumps within the same base version (`1.15.3-beta.1` -> `1.15.3-beta.2`), never across (`-> 1.18.0-beta.1`) — so it drifts and has to be realigned by hand, which in practice never happens. The beta stream tracks the stable one 1:1, so allowing prereleases on this package alone, inside the OpenTelemetry group, keeps the four versions in step automatically.

Deliberately scoped to the Prometheus exporter only: the stable stream also publishes `rc` versions (`1.16.0-rc.1`, `1.18.0-rc.1`) that must not reach production packages.

# Testing

No code change, so no test added.

- `renovate-config-validator` passes (exit 0).
- The rules were replayed against the 64 nuget packages actually referenced by the repository's `.csproj` files, to confirm the resulting grouping: 25 update streams, `ArmoniK.*` and `Htc.Mock` correctly staying on the immediate cadence, and `schedule` / `bumpVersion` correctly inherited from the catch-all rule by every new group.

# Impact

- **More PRs, each smaller**: 25 update streams instead of 1 group. These are streams, not simultaneous PRs — one only appears in a month where its package actually publishes. Expect a burst on the first run, since 7 months of pending updates come out at once.
- **CI cost increases**, as each PR runs the full matrix. `prConcurrentLimit` can throttle the burst if needed; `config:base` already caps PR creation at 2/hour.
- **#1159 and #402 are superseded** by the split PRs and should be closed.
- The two known-failing updates become isolated: NUnit and MongoDB.Driver each land in their own PR with a documented cause, instead of blocking 38 others.
- **No behaviour change for `bumpVersion: "patch"`** — it is inherited from the catch-all rule by all new groups.

# Additional Information

`renovate-config-validator` reports an available config migration, but it concerns only pre-existing fields (`config:base` -> `config:recommended`, `matchPackagePrefixes` / `excludePackagePrefixes` -> `matchPackageNames`). None of the rules added here appear in the migration diff. Left out of scope — worth a separate cleanup.

Two follow-ups this PR does **not** address, both requiring real code changes:

- The 37 `[Timeout(n)]` in `Common/tests` (5 files) to migrate to `[CancelAfter(n)]`. Note it works by cooperative cancellation, so tests that must actually be interrupted need the `CancellationToken` threaded through.
- The MongoDB.Driver / PostgresEmbed conflict, which exists only because `Adaptors/PostgresSQL/tests` transitively pulls MongoDB.Driver through `Common/tests` in order to reach 8 abstract base classes.

# Checklist


renovate.json
```diff
@@ -31,7 +31,122 @@
...
+      "groupSlug": "nuget-otel",
+      "matchDatasources": [
+        "nuget"
+      ],
+      "matchPackageNames": [
+        "OpenTelemetry*"
+      ]
+    },
+    {
+      "groupName": "OpenTelemetry",
```
jfonseca-aneo, line 92: should't `prometheus-net` be added to this group as well ?

renovate.json
```diff
@@ -31,7 +31,122 @@
...
+    },
+    {
+      "groupName": null,
+      "matchDatasources": [
+        "nuget"
+      ],
+      "matchPackageNames": [
+        "MongoDB.*",
+        "Npgsql",
+        "StackExchange.Redis",
```
jfonseca-aneo, line 141: ` redis-inside ` ?

renovate.json
aneojgurhem: No this group is here to manage the fact that one of the Opentelemetry package is in beta.

renovate.json
aneojgurhem: Good catch — adding it.

`redis-inside` 3.3.0 is referenced by `Adaptors/Redis/tests` and `Adaptors/Embed/tests`, and it is an embedded test server: the same category as `EphemeralMongo` and `MysticMind.PostgresEmbed`, which are both already in this list.

It also has the exact property that motivates the list: **0 stable releases in the last 12 months**. Isolating it costs nothing today (it will not open a single PR), but a package dormant for years is precisely the one that lands a large version jump when it finally moves. That is literally what happened on #1159 with `MysticMind.PostgresEmbed`, whose SharpCompress 0.34.2 requirement collided with the 0.48.1 floor raised by MongoDB.Driver 3.9.0 — 267 test failures. Left in the catch-all group, that kind of breakage blocks every other update again, which is the failure mode this PR exists to remove.

jfonseca-aneo (approved):

